PR #34, opened 2026-01-23
fix(bug-finding): improve build pipeline robustness

0xdkay:
## Summary

- Fix subprocess hangs by passing `stdin=DEVNULL` to all subprocess calls
- Filter docker exit code 0 false errors from stderr output
- Fix signal handler re-entry race condition during cleanup
- Skip LiteLLM/Postgres deployment when `model_list` is empty
- Add `--no-cache` flag to `prepare` and `build` commands (disables docker layer caching)
- Update documentation (README, INTEGRATION.md, docs/README.md) for factual accuracy

## Test plan

- [x] Tested prepare, build, and run with `atlantis-multilang-given_fuzzer`

### DIFF
--- a/bug_finding/src/run.py
+++ b/bug_finding/src/run.py
@@ -10,8 +10,6 @@ from collections.abc import Callable
 from importlib.resources import files
 from pathlib import Path
 
-import yaml
-
 from bug_finding.src.render_compose import render as render_compose
 from bug_finding.src.crs_utils import (
     clone_oss_fuzz_if_needed,
@@ -27,26 +25,6 @@ logger = logging.getLogger(__name__)
 # Default registry path
 # __package__ is guaranteed to be set when running as a module
 DEFAULT_REGISTRY_DIR = files(__package__).parent.parent / "crs_registry"  # type: ignore[arg-type]
-
-
-def _needs_litellm(config_dir: Path) -> bool:
-    """Check if LiteLLM is needed based on config-litellm.yaml.
-
-    Returns False if model_list is empty/None (CRS doesn't use LLM).
-    """
-    litellm_config = config_dir / "config-litellm.yaml"
-    if not litellm_config.exists():
-        return False
-    try:
-        with open(litellm_config) as f:
-            config = yaml.safe_load(f)
-        if config is None:
-            return False
-        model_list = config.get("model_list")
-        # Empty list or None means no models configured
-        return bool(model_list)
-    except (yaml.YAMLError, AttributeError, TypeError):
-        return False
 
 
 def run_crs(
@@ -67,6 +45,7 @@ def run_crs(
     harness_source: Path | None = None,
     diff_path: Path | None = None,
     external_litellm: bool = False,
+    skip_litellm: bool = False,
     source_oss_fuzz_dir: Path | None = None,
     ensemble_dir: Path | None = None,
     disable_ensemble: bool = False,
@@ -95,6 +74,7 @@ def run_crs(
         harness_source: Optional path to harness source file (Path, already resolved)
         diff_path: Optional path to diff file (Path, already resolved)
         external_litellm: Use external LiteLLM instance (default: False)
+        skip_litellm: Skip LiteLLM/Postgres deployment entirely (default: False)
         source_oss_fuzz_dir: Optional source OSS-Fuzz directory to copy from (Path, already resolved)
         ensemble_dir: Optional base directory for ensemble sharing (Path, already resolved)
         disable_ensemble: Disable automatic ensemble directory for multi-CRS mode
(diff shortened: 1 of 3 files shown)


bug_finding/src/run.py
```diff
@@ -27,6 +29,26 @@
 DEFAULT_REGISTRY_DIR = files(__package__).parent.parent / "crs_registry"  # type: ignore[arg-type]
 
 
+def _needs_litellm(config_dir: Path) -> bool:
+    """Check if LiteLLM is needed based on config-litellm.yaml.
+
+    Returns False if model_list is empty/None (CRS doesn't use LLM).
+    """
+    litellm_config = config_dir / "config-litellm.yaml"
```
azchin: This conflicts with #35, where I hope to make `config-litellm.yaml` optional (as we discussed...). This is to simplify the user config interface.

A better place you could check is `config-crs.yaml:models`, but if that's empty then the default behaviour is to enable all models. But I'm considering deprecating the `models` field because it might be yet another obstacle to easy CRS integration. We need to discuss whether to:
1. Keep `models` and use it to detect whether to skip LiteLLM
2. Deprecate `models` for simpler CRS integration interface

Personally, I think fuzzers w/out LLMs should just accept the startup overhead of LiteLLM, or use `--external-litellm` with a dummy key.

bug_finding/src/prepare.py
```diff
@@ -152,10 +153,14 @@ def build_bake_images(crs_path: Path) -> bool:
         return False
 
     logger.info("Building images with docker buildx bake in %s", crs_path)
+    cmd = ["docker", "buildx", "bake", "-f", str(bake_file)]
```
azchin, line 156: Just curious, why is disabling cache a desired feature?

bug_finding/src/build.py
```diff
@@ -506,11 +512,12 @@ def build_crs(
                     str(compose_file),
                     "--profile",
                     profile,
-                    "up",
-                    "--abort-on-container-exit",
+                    "run",
```
azchin, line 506: I see, artifact from when we needed to start the key provisioner along the builder.
Since you're running each service manually, can you just get rid of profiles altogether in `compose.yaml.j2`?

bug_finding/src/run.py
0xdkay: > I hope to make config-litellm.yaml optional.

I agree with this. Lets go for it.

> Personally, I think fuzzers w/out LLMs should just accept the startup overhead of LiteLLM, or use --external-litellm with a dummy key.

If that is the case, we can make something like `--no-litellm` explicitly for the CRS that runs only fuzzers. (only if second topic is decided not to skip it automatically).

For the second one, I am still not sure. Specifying models in `config-crs.yaml` has any benefits from the `oss-crs` side?

bug_finding/src/prepare.py
0xdkay: When developing, because of cache issues, some of the changes may not be reflected in the docker image. One typical example is if they run a custom builder script, it will just keep the docker container as-is, without running the build script again.

This issue might be related to the fact that the build phase only accepts a single Dockerfile. If we accept something like docker-compose file, would it be better? (is it possible?)

We can provide `--force-rebuild` which clears existing build and out dir for a crs insetad of `--no-cache`.

bug_finding/src/run.py
azchin: > I agree with this. Lets go for it.

Thanks! I'll merge #35, or you can review it and block if you have comments/concerns.

> Specifying models in config-crs.yaml has any benefits from the oss-crs side?

Specifying models in `config-crs.yaml` unlocks the following features

- Cross checking against `config-litellm.yaml` to make sure user setup supports CRS requirements
- Restrict the provisioned LiteLLM virtual key to only the models specified by CRS

I forgot to mention, there's a third option which is to keep the current interface: if `models` is provided, then we restrict models. If not or it's empty, we allow all models. _But it turns out this wasn't implemented correctly :P_, which is why I don't mind changing the interface now.

> If that is the case, we can make something like --no-litellm explicitly for the CRS that runs only fuzzers. (only if second topic is decided not to skip it automatically).

Hm feels like interface bloat. How about checking `config-resource.yaml` for 0 LLM budget?

bug_finding/src/prepare.py
azchin: I added a `oss-bugfind-crs clean` command that does the clearing existing build and out dir. I'm fine with `--no-cache`.

bug_finding/src/build.py
0xdkay: Fixed via https://github.com/sslab-gatech/oss-crs/pull/34/commits/6a9bea3d3ccabe439ba0d4eb82396dec195c7936

bug_finding/src/run.py
0xdkay: As we discussed, `0` can also mean `unlimited`, and let users choose this, lets add `--skip-litellm` explicitly.

Done via b6281df0da1710d12e916dbaf6c57a669957911c

PR #59, opened 2026-02-09
docs: RFC

azchin:
RFC for the OSS-CRS interface, configurations, and CRS integration.

We appreciate any and all feedback! Leave a comment or review in this issue if you are interested in discussing.


HQ1995:
One major consideration is whether we should combine bug-finding and bug-fixing into a single RFC, as we are merging these two processes together.

docs/RFC.md
```diff
@@ -0,0 +1,612 @@
...
+- [Rollout / Timeline](#rollout--timeline)
+- [References](#references)
+
+---
+
+## Overview and context
+
+This RFC proposes a standardized set of interfaces and specifications for Cyber Reasoning Systems (CRSs) following the DARPA AIxCC competition.
+The aim is to unify bug finding and reporting workflows to make them portable across teams, scalable from laptops to clusters, and compatible with oss-fuzz.
+
```
david-a-wheeler, line 40: I'm sure that's the short-term aim, but this overview text doesn't justify why anyone would *care* about unifying bug finding and reporting workflows. This overview should also give the "bigger picture" that I think you know but aren't expressing. I suggest adding something like this:

```suggestion
This unification enables CRSs to interoperate and accurately measure their capabilities, leading to steady improvements over time so that real-world software vulnerabilities can be automatically identified and fixed at scale, vastly reducing the risks to society from vulnerable software.

```

david-a-wheeler: (If you want, you can accept this comment, but feel to write something else. My point is that you need to explain why this can help users and society, not just what it does, to put it in context.)

docs/RFC.md
```diff
@@ -0,0 +1,612 @@
...
+This RFC proposes to standardise the following systems:
+
+- Bug Finding Systems
+- Tools for LLM-based systems
+
+**The implementation of this proposal: OSS-CRS ([https://github.com/sslab-gatech/oss-crs/tree/main](https://github.com/sslab-gatech/oss-crs/tree/main)) is an independent standard CRS framework that can test and run against OSS-Fuzz projects.**
+
+## Glossary and terms
```
david-a-wheeler, line 49: I know what these are, but does the *reader* know? These are just acronym expansions, which is a good start, but you might want to follow each with a 1-sentence definition if you think some readers might not know what these are (as you already did with libCRS).

docs/RFC.md
```diff
@@ -0,0 +1,612 @@
...
+#### Highlighting the Impact of LLMs through Fine-Grained Bug-Finding Phase Control
+
+In most existing OSS-Fuzz use cases such as FuzzBench, fuzzing starts from scratch and runs until coverage stops growing. It usually takes a long time before we can tell which fuzzer or setup performs better, often by comparing the total number of bugs found or the final coverage.
+
+In this RFC, we want to make this process more flexible.
+By adding fine-grained control of fuzzing phases, we can run AI agents at different stages such as the early exploration phase or the later saturation phase.
+This helps us understand where and how much LLMs can contribute without waiting for the entire campaign to finish.
+
+#### Pushing OSS-Fuzz Closer to developers via Diff Mode
+
```
david-a-wheeler, line 84: Nowhere does this explain what "Diff Mode" is. I mean, I can guess, but if that's a key concept, it should be briefly defined in this section (since this section is all about Diff Mode).

docs/RFC.md
```diff
@@ -0,0 +1,612 @@
...
+- Support both oss-fuzz upstream integration and local runs.
+
+### Non-Goals
+
+- Define new vulnerability discovery algorithms.
+- Replace oss-fuzz; instead, extend compatibility.
+
+## Background
+
+- **CRS (Cyber Reasoning System):** automation systems that find and patch vulnerabilities.
```
david-a-wheeler, line 118: Shouldn't this be in the definitions/glossary earlier, so you don't need to repeat it here?

docs/RFC.md
```diff
@@ -0,0 +1,612 @@
...
+- **AIxCC experience:** each team had strong but siloed implementations. Standardization is the natural next step.
+- **Motivating pain points:**
+  - Lack of consistent interface for invoking bug finding vs patching.
+  - No standardized format for benchmarks (POVs, metadata).
+  - Inconsistent patch aggregation policies.
+
+## Standardization Proposals / CRS Integration Topology
+
+### Command Structure
+
```
david-a-wheeler, line 128: Add some intro text, e.g.:

```suggestion

Here are examples of command-line commands that our work would enable:
```

docs/RFC.md
```diff
@@ -0,0 +1,612 @@
...
+| `source.local_path` | string | Yes* | Alternative: local path |
+| `cpuset` | string | Yes | CPU cores (e.g., `"4-7"`, `"0,2,4,6"`) |
+| `memory` | string | Yes | Memory limit (e.g., `"16G"`) |
+| `llm_budget` | integer | No | LLM budget in dollars |
+
+\* Either `url` + `ref` OR `local_path` must be provided.
+
+### LLM Configuration
+
+LLM configuration uses the native [LiteLLM Proxy Configuration](https://docs.litellm.ai/docs/proxy/configs) format. The configuration file path is specified in the CRS-Compose file via `llm_config.litellm_config`.
```
david-a-wheeler, line 251: You suddenly introduce litellm here, without any fanfare, yet it's implicitly defining key parts of the interface.

I don't *object* to the use of litellm, but since it's embedded in the proposal, you really should introduce litellm earlier & why you're using it. E.g., "As part of this proposal we propose embedding the litellm Python library...." Explain briefly why you're using it, why not some other library. Then it won't be a surprise later on.

docs/RFC.md
```diff
@@ -0,0 +1,612 @@
...
+```
+
+#### LLM Availability Validation
+
+CRSs declare their LLM requirements via `required_llms` in `oss-crs/crs.yaml`. At runtime, oss-crs validates that all required models are available in the LiteLLM configuration before launching the CRS.
+
+### Diff Mode Support
+
+Diff Mode lets the CRS focus on code changes between commits instead of fuzzing the whole project.
+It's useful for validating patches, PRs, or a series of commits where you only want to explore the modified code paths.
```
david-a-wheeler: That's a fine explanation of Diff Mode, except you already introduced Diff Mode earlier. You should explain what Diff Mode is/does the first time you use the term. Maybe that goes early in the document, to make sure readers are aware of it. Diff Mode seems like a key new capability that deserves early discussion.

docs/RFC.md
```diff
@@ -0,0 +1,612 @@
...
+uv run oss-crs run \
+  --compose-file <co<path> \
+  --focus-diff <path> \
+  --target-proj-path <path> \
+  --target-harness <harness>
+```
+
+### CRS Ensemble Support
+
+**Motivation:** Combine pure agents with fuzzers for maximum coverage.
```
david-a-wheeler: I'm not sure what a "pure" agent is. Is this only using AI? Only doing static analysis?

I'm all for combining and using fuzzers, but I think we should use all tools available for finding vulnerabilities - if a traditional static analysis tool, or an AI reading code, can help find vulnerabilities, great! Indeed, I'd love to see human inputs as a potential source (e.g., "I'm especially concerned that the interaction of X and Y might produce Z, can you look into that?").

docs/RFC.md
```diff
@@ -0,0 +1,612 @@
...
+crs.submit(DataType.POV, file_path)            # Submit a single file
+crs.fetch(DataType.SEED, directory)            # Fetch shared data
+
+# Service Discovery
+domain = crs.get_service_domain("litellm")     # Get service endpoint
+```
+
+#### Environment Variables
+
+CRSs receive the following environment variables:
```
david-a-wheeler:
```suggestion
CRSs may receive the following environment variables:
```

Are all of these strictly required? I doubt it. I suggest that these be as optional as you can make them.

docs/RFC.md
```diff
@@ -0,0 +1,612 @@
...
+| atlantis-multilang | bug-finding | LLM-powered multi-language fuzzer |
+
+## Risks / Open Questions
+
+- Multi-language CRS support (C, Rust, Java).
+- Distributed System support.
+- Secure handling of LLM API keys.
+- Resource isolation between CRS instances.
+
+## References
```
david-a-wheeler, line 793: I would expect the AIxCC competition (at least some key docs) be included in the references. What about the oss-fuzz documentation?

david-a-wheeler (commented):
Here are several comments, I hope they help!

PR #62, opened 2026-02-16
feat: implement libCRS with builder & exchange sidecars and data pre-population (--pov, --diff, --corpus)

### LINKED ISSUE #60: Incremental build support for bug-fixing CRS
## Summary

Enable bug-fixing CRSs to perform incremental rebuilds using Docker snapshot images. oss-crs creates a **target-level snapshot** of the compiled project (opt-in via config). CRS developers can use this snapshot directly or extend it with custom tools via the **builder sidecar pattern**. Communication between agent and builder goes through libCRS's `apply-patch-build` API — backend-agnostic, same CRS code works on local Docker Compose and future cloud deployments.

## Approach

Follows the **oss-fuzz Chronos** approach (uses `make_build_replayable.py`, `replay_build.sh`, `REPLAY_ENABLED` from base-builder):
- `docker commit -c 'ENV REPLAY_ENABLED=1'` after compilation → target-level snapshot
- `make_build_replayable.py` runs lazily during first replay build (already in base-builder image)
- Handler execution order: `git apply` (real binary, before stubbing) → `compile` (triggers replay build) → optional verify/test
- Snapshot created via raw `docker run` (not docker compose, not `--rm`), container survives for commit

### Key Design Decisions

1. **Target-level snapshot (NOT CRS-level)** — compilation is project-level, one snapshot per target, shared across CRS modules
2. **Builder sidecar pattern** — CRS developers extend snapshot with custom tools via `FROM $snapshot_image`
3. **Ephemeral execution (local)** — `docker run --rm` per request, clean state, enables concurrent builds
4. **Single container per request** — apply patch + compile + verify + test all in one `docker run`
5. **Image-agnostic** — CRS specifies which snapshot image to use via `image` parameter
6. **Snapshot naming** — `{name}:{repo_hash}-{sanitizer}-snapshot` (includes sanitizer)
7. **Volume-mount `/out`** during snapshot creation to exclude fuzzer binaries
8. **Lazy `make_build_replayable.py`** — handler calls `git apply` BEFORE `compile`, so real git binary is used for patching. When `compile` subsequently triggers stubbing, patching is already done.

### Phase 1: Target-level snapshot creation and compose template
- Run `compile` on target base image via raw `docker run` (not `--rm`)
- Copy `replay_build.sh` to `$SRC/`, bake default handler script
- `docker commit -c 'ENV REPLAY_ENABLED=1'` → snapshot image
- Volume-mount `/out` to exclude fuzzer binaries
- Pass `OSS_CRS_SNAPSHOT_IMAGE` env var and `snapshot_image` build arg to CRS modules
- Add `incremental_build: true` config flag
- Clean up build container after commit (`docker rm`)

### Phase 2: libCRS apply-patch-build (ephemeral mode, image-agnostic)
- Add `apply_patch_build(patch_path, dst_path, timeout, image, verify_pov, run_tests)` to CRSUtils
- Implement `LocalCRSUtils.apply_patch_build()` — ephemeral `docker run --rm`
- Define BuildRequest/BuildResult data models
- Add `apply-patch-build` CLI command
- Docker socket handled by oss-crs infrastructure, not CRS modules
- Single container per request to avoid multiple bootup overhead

### Phase 3: Documentation and example builder module
- Document apply-patch-build interface in `docs/design/libCRS.md`
- Document incremental build pattern and builder sidecar in `docs/crs-development-guide.md`
- Provide example handler script + Dockerfile

### Snapshot Creation (bypasses docker compose)

The existing `docker compose run --rm` flow destroys containers before we can `docker commit`. Snapshot creation uses a separate path:

```python
# Raw docker run (no --rm) → compile → commit → cleanup
docker run --name {container} -v {out_dir}:/out {target_base_image} \
    bash -c "compile && cp /usr/local/bin/replay_build.sh $SRC/"
docker commit -c 'ENV REPLAY_ENABLED=1' {container} {snapshot_tag}
docker rm {container}
```

### Handler Execution Order (critical)
```
1. git apply patch.diff    — BEFORE compile, real git binary is used
2. compile                 — triggers replay_build.sh → make_build_replayable.py
                             (stubs git/make/cmake) → build.sh (replay)
3. verify pov (optional)   — run fuzzer binary against POV
4. run tests (optional)    — run_tests.sh (with RTS if enabled)
```

Note: `make_build_replayable.py` stubs `git apply` as no-op and replaces `patch` with dummy. But patching already happened in step 1. If post-stub git is ever needed, `/usr/bin/git.real` has the real binary. Handler should be resilient: try `git apply`, fall back to `git.real apply`.

### Snapshot Layering Options
- **Option A (simple)**: target-level snapshot → handler script → ephemeral
- **Option B (custom)**: target-level snapshot → CRS tools → CRS snapshot → handler → ephemeral
- **Option C (with RTS)**: target-level snapshot (with RTS baked in) → handler → ephemeral

### Cloud Deployment Model

In cloud environments (Azure ACI, Kubernetes), ephemeral containers have significant overhead (10-30s startup, image pull, scheduling). Cloud operators would use a **persistent sidecar** model:

```
K8s pod:
  ┌──────────────┐  shared volume  ┌──────────────────┐
  │ Agent         │ ◄════════════► │ Builder sidecar   │
  │ container     │  request/resp  │ (from snapshot)   │
  │               │                │ long-running:     │
  │ libCRS API    │                │   accept request  │
  │               │                │   git apply       │
  │               │                │   compile         │
  │               │                │   reset state     │
  │               │                │   repeat          │
  └──────────────┘                └──────────────────┘
```

The `apply_patch_build()` API stays the same — ephemeral vs persistent is a backend concern:
- **Local (`LocalCRSUtils`)**: spawns ephemeral `docker run --rm` per request
- **Cloud (`CloudCRSUtils`)**: communicates with persistent sidecar via shared volume or HTTP

Snapshot creation in cloud can use a Dockerfile-based build (`RUN compile && ... && rm -rf /out/*`) instead of `docker commit`, compatible with any container build service (ACR Tasks, Cloud Build, Kaniko).

## Future Work (TODO)

- [ ] **Persistent builder mode (cloud deployment)**: Keep builder sidecar alive across requests. Required for cloud environments where container startup is expensive. Adds state reset protocol, shared volume or HTTP communication. This is the expected cloud deployment pattern, not just a nice-to-have.
- [ ] **JVM incremental build**: Uses `maven-build-cache-extension` (NOT `REPLAY_ENABLED`) — fundamentally different mechanism from C/C++. Requires `extensions.xml` in `.mvn/` directory.
- [ ] **ccache integration**: Volume-mount ccache during snapshot creation to exclude from snapshot but speed up builds (Chronos pattern: `-v=ccache:/workspace/ccache`)
- [ ] **Dockerfile-based snapshot creation**: Replace `docker commit` with `RUN compile` in a Dockerfile for cloud-native builds

## References

- oss-fuzz Chronos: `infra/chronos/` in oss-fuzz repo
- oss-fuzz base-builder: `infra/base-images/base-builder/` (contains `make_build_replayable.py`, `replay_build.sh`)
- Planning docs: `.planning/` directory

0xdkay:
## Summary

- **Incremental build system**: Snapshot-based builds for oss-fuzz targets via an HTTP API builder sidecar that handles build, run-pov, and run-test requests
- **Builder as first-class CRS**: Standalone CRS type with its own compose config, Dockerfile, and dedicated entry point
- **Artifact isolation & shared dirs**: Per-build artifact directories with libCRS shared directories for cross-container transfer; oss-fuzz reproduce integration for POV validation
- **Robustness**: Exponential-backoff health polling at startup, build dedup cache that doesn't persist failures, pinned oss-fuzz commit, absolute `git.real` path for patch application
- **Registry**: Add `crs-claude-code` to the CRS registry

## Test results

Tested with `crs-claude-code` by running `crs-compose run` against `sanity-mock-c-delta-01` with builder-enabled compose:
- [x] Snapshot build caching (second build reuses cached image)
- [x] POV replay via `/run-pov` endpoint
- [x] Test execution via `/run-test` endpoint
- [x] Health polling prevents startup race conditions

Closes #60

### DIFF
--- a/docs/crs-development-guide.md
+++ b/docs/crs-development-guide.md
@@ -297,7 +297,7 @@ Your containers receive these environment variables automatically:
 | `OSS_CRS_BUILD_OUT_DIR` | Build output directory (read-only at run time) | `/OSS_CRS_BUILD_OUT_DIR` |
 | `OSS_CRS_SUBMIT_DIR` | Submission directory | `/OSS_CRS_SUBMIT_DIR` |
 | `OSS_CRS_SHARED_DIR` | Shared directory (between containers in this CRS) | `/OSS_CRS_SHARED_DIR` |
-| `OSS_CRS_EXCHANGE_DIR` | Inter-CRS data exchange + bootup data (shared across all CRSs) | `/OSS_CRS_EXCHANGE_DIR` |
+| `OSS_CRS_FETCH_DIR` | Inter-CRS data exchange + bootup data (read-only, shared across all CRSs) | `/OSS_CRS_FETCH_DIR` |
 | `FUZZING_ENGINE` | OSS-Fuzz fuzzing engine | `libfuzzer` |
 | `SANITIZER` | OSS-Fuzz sanitizer | `address` |
 | `ARCHITECTURE` | Target architecture | `x86_64` |
@@ -325,10 +325,10 @@ libCRS skip-build-output <dst_path>
 libCRS register-submit-dir <type> <path>      # type: pov, seed, bug-candidate, patch
 libCRS register-shared-dir <local_path> <shared_path>
 
-# Fetch directory registration (daemon watcher for EXCHANGE_DIR)
+# Fetch directory registration (daemon poller for FETCH_DIR)
 libCRS register-fetch-dir <type> <path>       # type: pov, seed, bug-candidate, patch, diff
 
-# One-shot fetch (copies from EXCHANGE_DIR)
+# One-shot fetch (copies from FETCH_DIR)
 libCRS fetch <type> <path>                    # type: pov, seed, bug-candidate, patch, diff
 
 # Manual submission
@@ -733,21 +733,21 @@ Your CRS should submit findings through libCRS:
 
 ## Fetching Data
 
-CRS containers receive data through `EXCHANGE_DIR`, a shared volume mounted to all non-builder containers. Data arrives from two sources:
+CRS containers receive data through `FETCH_DIR`, a read-only volume mounted to all non-builder containers. Data arrives from two sources:
 
-1. **Bootup data** — Files passed via `crs-compose run` flags (`--pov-dir`, `--diff`, `--corpus`), pre-populated into `EXCHANGE_DIR` before containers start.
-2. **Inter-CRS data** — Files submitted by other CRSs at runtime via `register-submit-dir` or `submit`.
+1. **Bootup data** — Files passed via `crs-compose run` flags (`--pov-dir`, `--diff`, `--corpus`), pre-populated by the host before containers start.
+2. **Inter-CRS data** — Files submitted by other CRSs at runtime via `register-submit-dir` or `submit`, delivered by the exchange sidecar which polls `SUBMIT_DIR` and copies artifacts into the shared exchange volume.
 
 ### Bootup Data (crs-compose run flags)
 
 The operator passes data via `crs-compose run`:
-- `--pov <file>` or `--pov-dir <dir>` — PoV files → `EXCHANGE_DIR/pov/`
-- `--diff <file>` — Reference diff → `EXCHANGE_DIR/diff/ref.diff`
-- `--corpus <dir>` — Seed corpus files → `EXCHANGE_DIR/seed/`
+- `--pov <file>` or `--pov-dir <dir>` — PoV files → `FETCH_DIR/pov/`
+- `--diff <file>` — Reference diff → `FETCH_DIR/diff/ref.diff`
+- `--corpus <dir>` — Seed corpus files → `FETCH_DIR/seed/`
 
-### Using register-fetch-dir (Daemon Watcher)
+### Using register-fetch-dir (Daemon Poller)
 
-For continuous data fetching, use `register-fetch-dir`. It forks a daemon that performs an initial sync and then watches `EXCHANGE_DIR` for new files:
+For continuous data fetching, use `register-fetch-dir`. It forks a daemon that performs an initial sync and then polls `FETCH_DIR` for new files:
 
 ```bash
 # In your entry script:
@@ -756,8 +756,8 @@ libCRS register-fetch-dir seed /my-seeds &
 ```
 
 The daemon:
-1. Copies existing files from `EXCHANGE_DIR/<type>/` into the local directory.
-2. Watches `EXCHANGE_DIR/<type>/` for new files and copies them as they arrive.
+1. Copies existing files from `FETCH_DIR/<type>/` into the local directory.
+2. Polls `FETCH_DIR/<type>/` periodically for new files and copies them as they arrive.
 
 ### Using fetch (One-Shot)
 
@@ -774,11 +774,11 @@ NEW_FILES=$(libCRS fetch pov /my-povs)
 
 ### Available Data Types
 
-| CLI Flag | Data Type | EXCHANGE_DIR Subdirectory |
+| CLI Flag | Data Type | FETCH_DIR Subdirectory |
 |---|---|---|
-| `--pov`, `--pov-dir` | `pov` | `EXCHANGE_DIR/pov/` |
-| `--diff` | `diff` | `EXCHANGE_DIR/diff/` |
-| `--corpus` | `seed` | `EXCHANGE_DIR/seed/` |
+| `--pov`, `--pov-dir` | `pov` | `FETCH_DIR/pov/` |
+| `--diff` | `diff` | `FETCH_DIR/diff/` |
+| `--corpus` | `seed` | `FETCH_DIR/seed/` |
 
 ---
 
--- a/PLAN.md
+++ b/PLAN.md
@@ -122,7 +122,7 @@ Features that will significantly improve the platform's utility for real-world c
 
 | Item | Status | Notes |
 |---|---|---|
-| `InfraClient` in libCRS | Stub only | `submit_batch` is currently a no-op; will connect to seed/PoV dedup services |
+| `InfraClient` in libCRS | ✅ Implemented | Exchange sidecar handles SUBMIT_DIR → EXCHANGE_DIR; InfraClient fetches from FETCH_DIR (read-only) |
 
 ---
 
--- a/crs/src/cli/crs_compose.py
+++ b/crs/src/cli/crs_compose.py
@@ -93,25 +93,25 @@ def add_run_command(subparsers):
         "--pov",
         type=Path,
         default=None,
-        help="Single POV file to copy into EXCHANGE_DIR",
+        help="Single POV file to pre-populate into FETCH_DIR before containers start",
     )
     run.add_argument(
         "--pov-dir",
         type=Path,
         default=None,
-        help="Directory containing POV files to copy into EXCHANGE_DIR",
+        help="Directory containing POV files to pre-populate into FETCH_DIR before containers start",
     )
     run.add_argument(
         "--diff",
         type=Path,
         default=None,
-        help="Path to a diff file for delta-mode analysis. Accessible via: libCRS register-fetch-dir diff <local_path>",
+        help="Diff file for delta-mode analysis, pre-populated into FETCH_DIR before containers start. Accessible via: libCRS fetch diff <local_path>",
     )
     run.add_argument(
         "--corpus",
         type=Path,
         default=None,
-        help="Directory containing initial corpus/seed files. Accessible via: libCRS register-fetch-dir seed <local_path>",
+        help="Directory of initial corpus/seed files, pre-populated into FETCH_DIR before containers start. Accessible via: libCRS fetch seed <local_path>",
     )
 
 
--- a/crs/src/crs.py
+++ b/crs/src/crs.py
@@ -266,6 +266,9 @@ class CRS:
 
         Navigates up from per-CRS work_dir (work_dir/crs/<name>) to compose
         work_dir (work_dir), then into EXCHANGE_DIR/<target>/<harness>/.
+
+        CRS containers mount this as FETCH_DIR (read-only). The exchange
+        sidecar and host-side pre-population are the only writers.
         """
         target_image_name = target.get_docker_image_name().replace(":", "_")
         assert target.target_harness, (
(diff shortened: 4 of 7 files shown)


crs/src/target.py
```diff
@@ -243,3 +289,199 @@ def get_target_env(self) -> dict:
...
+            setup_container_name = f"{container_name}-setup"
+            setup_cmd = [
+                "docker", "run",
+                f"--name={setup_container_name}",
+                "--privileged",
+                intermediate_tag,
+                "bash", "-c",
+                # chmod all scripts in /usr/local/bin/ that we copied
+                "chmod +x /usr/local/bin/oss_crs_handler.sh "
+                "/usr/local/bin/compile "
```
azchin: Programatically construct with `OSS_FUZZ_SCRIPTS.keys()`?

scripts/setup-third-party.sh
azchin, line 1: Does the user have to run this script to fully set up oss-crs? I think git submodule is better if user needs to manually call this (if submodule has similar sparse checkout mechanism).

Why not automatically call this script in oss-crs at runtime? Is it to make it explicit to the user that multiple licenses are involved?

And do we need NOTICE file? I don't see one in this PR.

crs/src/templates/oss_crs_builder_server.py
azchin, line 1: Currently, we have one builder using oss-crs-infra's resource specification, which hosts web-server that runs the `compile` inside its own container.

I was thinking keeping the web-server separate from the builder, and dynamically launch `<CRS>_builder_<id>` on host, using same cpuset as CRS. The web-server would then be hosted in oss-crs code instead of oss_crs_infra container.

0xdkay: how do you manage the web server to access to builder ?

scripts/setup-third-party.sh
0xdkay: submodule also has the sparse checkout mechanism. we do not need NOTICE file AFAIK.
The ones who needs builder sidecar can run this, or we can also make it automatically run if not installed.

crs/src/target.py
0xdkay: That makes sense.

crs/src/templates/oss_crs_builder_server.py
azchin: Web-server would not be run inside a container, but rather as another thread of the root OSS-CRS process. Web-server dispatches builder by doing `docker run` (or reuse compose file?).

Is there anything web-server needs from builder other than final artifacts and exit code?

scripts/setup-third-party.sh
azchin: Run automatically sounds good.

crs/src/templates/oss_crs_builder_server.py
0xdkay: I think that concern is ephemeral vs persistent builders.

I think persistent builder is better, as it does not require additional docker commands, and it does not need an extra startup time. Also, currently we put the build outputs to the host system, but we can remove those files after terminating the container.

crs/src/target.py
0xdkay: Completed via [c259c51](https://github.com/sslab-gatech/oss-crs/pull/62/commits/c259c51654cd818d6f746fccda335c6764f1eb7c)

scripts/setup-third-party.sh
0xdkay: Completed via [c259c51](https://github.com/sslab-gatech/oss-crs/pull/62/commits/c259c51654cd818d6f746fccda335c6764f1eb7c)

docs/config/crs.md
```diff
@@ -152,6 +172,24 @@ crs_run_phase:
...
+
+```yaml
+crs_run_phase:
+  patcher:
+    dockerfile: oss-crs/docker-compose/patcher.Dockerfile
+    additional_env:
+      MAX_PATCHES: "100"
+  builder-asan:
+    dockerfile: oss-crs-infra:default-builder
+    use_snapshot: true
```
azchin: Is `use_snapshot` clear? Maybe `runtime_builder`, `sidecar_builder`?

docs/crs-development-guide.md
```diff
@@ -472,6 +490,181 @@ FUZZER_ADDR=$(libCRS get-service-domain fuzzer)
...
+
+```bash
+#!/bin/bash
+# run-patcher.sh — entry script for a patcher module
+
+# 1. Generate a patch (your CRS logic)
+generate_patch > /tmp/patch.diff
+
+# 2. Apply the patch and rebuild (specify which builder sidecar to use)
+libCRS apply-patch-build /tmp/patch.diff /tmp/build-result --builder builder-asan
```
azchin, line 596: This is blocking til build finishes? Good stuff.

builder/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+## Input Validation
+
+The builder validates all user-provided parameters:
+- `harness_name`: alphanumeric, hyphens, underscores, or dots only
+- `build_id`: alphanumeric, hyphens, or underscores only
+
+## Build Artifacts
+
+Build artifacts from each patch are stored at `/builds/{build_id}/out/` inside the builder container. These are shared with other CRSs via `libCRS register-shared-dir`. A base (unpatched) build is created automatically at startup as `/builds/base/out/`.
```
azchin: This means on host it's under SHARED_DIR, right? Works for me. I'm working on a `build-id` of my own, but that's for host directories, not exposed to the containers.

docs/config/crs.md
0xdkay: `use_snapshot` describes the mechanism (this module uses the snapshot Docker image as its base), not the intent (this module is a builder sidecar). `use_snapshot: true` means "start from the snapshot image built during target_build_phase". The alternative names like `sidecar_builder` would be misleading because `use_snapshot` could theoretically be used for non-builder modules too (any module that needs the pre-built snapshot).

docs/crs-development-guide.md
0xdkay: Yes. libCRS internally handles polling but provides blocking to the crs developers.

builder/README.md
0xdkay: Yes. `SHARED_DIR/builds/{build_id}/out/`.  Slightly updated https://github.com/sslab-gatech/oss-crs/pull/62/commits/db739d0a24b79c75845fe088db8eb0fe30122430

docs/config/crs.md
azchin: Personally I prefer using intent over mechanism. But point taken about the theoretical case.

"use" is still ambiguous to me, use for what? `run_snapshot`, `runner_snapshot`?

0xdkay: run seems better

0xdkay: Completed via https://github.com/sslab-gatech/oss-crs/pull/62/commits/ccc9ae30a986dd4ec227aa4d0f02aca6bb40827a

libCRS/libCRS/cli/main.py
```diff
@@ -4,15 +4,16 @@
 from pathlib import Path
 from ..base import DataType, CRSUtils
 from ..local import LocalCRSUtils
-from ..common import OSS_CRS_RUN_ENV_TYPE, EnvType
+from ..common import get_run_env_type, EnvType
 
 
 def init_crs_utils() -> CRSUtils:
-    if OSS_CRS_RUN_ENV_TYPE == EnvType.LOCAL:
+    env_type = get_run_env_type()
```
azchin, line 11: Interesting... in the Dockerfile, how is this triggered? Just curious about the issue

crs/src/templates/run-crs-compose.docker-compose.yaml.j2
```diff
@@ -131,6 +158,7 @@ services:
           - postgres.oss-crs-infra-only
   oss-crs-litellm-key-gen:
     image: oss-crs-litellm-key-gen:latest
+    attach: false
```
azchin, line 163: FYI in the old implementation, I had this container sleep indefinitely after one-shot. This approach `attach: false` looks good to me.

crs/src/templates/run-crs-compose.docker-compose.yaml.j2
```diff
@@ -72,13 +84,27 @@ services:
     volumes:
       - {{ crs.get_build_output_dir(target) }}:/OSS_CRS_BUILD_OUT_DIR:ro
       - {{ crs.get_submit_dir(target) }}:/OSS_CRS_SUBMIT_DIR:rw
-      - {{ crs.get_fetch_dir(target) }}:/OSS_CRS_FETCH_DIR:ro
       - {{ crs.get_shared_dir(target) }}:/OSS_CRS_SHARED_DIR:rw
+{%- if not module_config.run_snapshot and not crs.config.is_builder %}
+      - {{ crs.get_exchange_dir(target) }}:/OSS_CRS_EXCHANGE_DIR:rw
```
azchin: So now each container libCRS directly uploads seeds / povs to a `rw` exchange dir?
Previously we had a sidecar that fetched things from the submit dir and then copied into exchange dir, but exchange dir was `ro` for CRS and `rw` for sidecar, I prefer for directory permissions.

What's the reasoning for doing exchange this way instead of the old sidecar way?

azchin (commented):
Main question about exchange dir perms.

crs/src/templates/run-crs-compose.docker-compose.yaml.j2
0xdkay: Ah, right. `OSS_CRS_EXCHANGE_DIR` should not be exposed to CRS as `rw`, and it should be handled outside of the container on the oss-crs's infra side.

libCRS/libCRS/cli/main.py
0xdkay:
```
  RUN pip3 install /libCRS \
      && python3 -c "from libCRS.base import DataType; print('libCRS OK')"
```

  `from libCRS.base import DataType` → loads `infra_client.py` → loads `common.py` → old code executes `OSS_CRS_RUN_ENV_TYPE = EnvType(get_env("OSS_CRS_RUN_ENV_TYPE"))` → `KeyError`

crs/src/templates/run-crs-compose.docker-compose.yaml.j2
0xdkay: Implemeted via https://github.com/sslab-gatech/oss-crs/pull/62/commits/db24951420585355d294fa17b7a37f8529ab700e

Issue #86, opened 2026-02-21
proposal: expose target build options (engine, sanitizer, architecture) to users


0xdkay:
## Current State

Two config files declare **supported options**:

- **project.yaml** (OSS-Fuzz format): `sanitizers`, `fuzzing_engines`, `architectures` — what the target project supports
- **crs.yaml**: CRS capabilities — what the CRS supports

But oss-crs only exposes `--sanitizer` to users. Engine and architecture are hardcoded to `libfuzzer` / `x86_64` in `get_target_env()`.

### 1. Expose all build options to users

Both engine and architecture should be user-selectable, not just sanitizer. The valid options come from the intersection of what the project supports and what the CRS supports.

### 2. Where to expose

Currently `--sanitizer` is only a CLI flag. These options could also be configurable in `compose.yaml`, so users can declare the build configuration declaratively.

### 3. Non-OSS-Fuzz targets

`project.yaml` is an OSS-Fuzz concept. If we support non-OSS-Fuzz targets, they won't have `project.yaml`. How do we handle option discovery and selection for those targets?

PR #87, opened 2026-02-21
fix(#79): pass sanitizer from CLI through to build and run templates

### LINKED ISSUE #79: fix: get_target_env() hardcodes sanitizer to 'address', ignoring project.yaml
## Bug

`Target.get_target_env()` in `oss_crs/src/target.py` hardcodes `"sanitizer": "address"` instead of reading from the target's `project.yaml`:

```python
def get_target_env(self) -> dict:
    # TODO: implement this properly
    ret = {
        "name": self.name,
        "language": self.config.language.value,
        "engine": "libfuzzer",        # ← also hardcoded
        "sanitizer": "address",        # ← hardcoded
        "architecture": "x86_64",
    }
```

This fallback is used by `CRSCompose._get_sanitizer()` when no `--sanitizer` flag is passed to `crs-compose build-target` or `crs-compose run`:

```python
@staticmethod
def _get_sanitizer(target: Target) -> str:
    return target.get_target_env().get("sanitizer", "address")
```

## Impact

Benchmarks that require a non-address sanitizer (e.g., `undefined` for UBSan) get silently built and run with `address` sanitizer when `--sanitizer` is not explicitly provided. This means:

- Bugs detectable only by UBSan (signed integer overflow, etc.) will never be caught even if the fuzzer triggers them
- The `sanitizers` field in `project.yaml` is effectively ignored

### Example

`afc-dav1d-full-01` has `sanitizers: [undefined]` in its `project.yaml` and its CPV is a signed integer overflow (CWE-190) requiring UBSan. Without `--sanitizer undefined`, crs-compose builds with ASan, making the bug undetectable.

## Suggested Fix

Read from `self.config` (which parses `project.yaml`) instead of hardcoding:

```python
def get_target_env(self) -> dict:
    sanitizer = self.config.sanitizers[0].value if self.config.sanitizers else "address"
    engine = self.config.fuzzing_engines[0].value if self.config.fuzzing_engines else "libfuzzer"
    return {
        "name": self.name,
        "language": self.config.language.value,
        "engine": engine,
        "sanitizer": sanitizer,
        "architecture": "x86_64",
    }
```

Note: `engine` is also hardcoded to `"libfuzzer"` — same issue.

## Workaround

Callers can pass `--sanitizer` explicitly to bypass the fallback. CRSBench will be updated to do this.

0xdkay:
## Summary

- `get_target_env()` hardcodes sanitizer to `"address"`, ignoring `--sanitizer` CLI flag
- Build and run templates used the hardcoded value, so `--sanitizer undefined` had no effect
- This caused dav1d (which requires UBSan) to be built with ASAN, missing the bug entirely

**Workaround**: thread the explicit `--sanitizer` value through to both templates, overriding `get_target_env()` at each call site. A proper fix (making `Target` hold the selected config) is deferred to #86.

## Verified

- `oss-crs build-target --sanitizer undefined` for dav1d: binary has 123 UBSan symbols, 0 ASAN symbols

Closes #79

## Test plan

- [x] `oss-crs build-target --sanitizer undefined` builds dav1d with UBSan
- [x] `oss-crs build-target --sanitizer address` still works for address sanitizer
- [x] Rendered build template contains `SANITIZER=undefined` when passed

### DIFF
--- a/oss_crs/src/crs_compose.py
+++ b/oss_crs/src/crs_compose.py
@@ -52,9 +52,7 @@ class CRSCompose:
 
     @staticmethod
     def _get_sanitizer(target: Target) -> str:
-        # Default sanitizer when --sanitizer is not provided via CLI.
-        # project.yaml lists *supported* sanitizers, not the selection.
-        return "address"
+        return target.get_target_env().get("sanitizer", "address")
 
     def set_deadline(self, deadline: float) -> None:
         self.deadline = deadline
--- a/oss_crs/src/target.py
+++ b/oss_crs/src/target.py
@@ -385,6 +385,7 @@ class Target:
         return None
 
     def get_target_env(self) -> dict:
+        # TODO: implement this properly
         # NOTE: engine/sanitizer/architecture are hardcoded defaults here.
         # project.yaml lists what the project *supports*, not what to use.
         # The actual selection should come from the user via CLI flags.


oss_crs/src/crs_compose.py
```diff
@@ -52,7 +52,9 @@ def _any_needs_snapshot(self) -> bool:
 
     @staticmethod
     def _get_sanitizer(target: Target) -> str:
-        return target.get_target_env().get("sanitizer", "address")
```
azchin, line 55: Should probably keep the `get_target_env` call so that we can implement that function in the future and have it pass through here?
(does this `_get_sanitizer` even get used?)

oss_crs/src/target.py
```diff
@@ -385,7 +385,11 @@ def _resolve_script_path(dst_name: str) -> Optional[Path]:
         return None
 
     def get_target_env(self) -> dict:
-        # TODO: implement this properly
```
azchin, line 388: If we want to do `project.yaml` parsing, this seems like a good place to do it? Don't remove TODO

oss_crs/src/crs_compose.py
0xdkay: We can handle this cleanly in #86

azchin (approved):
Ty sorry for holdup

PR #88, opened 2026-02-21
cgroup parent setup

### LINKED ISSUE #18: feat(bug-finding): run multiple docker containers without dind, via docker compose
Introduce a new interface, where CRS developers can provide a set of Docker containers the run directly on host (no need for DinD). We can let CRS developers provide a Docker compose YAML configuration.

Resource constraints can still be applied by using `--cgroup-parent` for memory limits, and the cpuset can stay the same (i.e. give the same cpuset to multiple containers when they belong to the same CRS).

Note this does not deprecate the DinD workflow, since this cannot handle container orchestration. All containers will be launched simultaneously from OSS-CRS via Docker compose.

azchin:
## Summary

This PR implements cgroup-parent support for shared resource limits across CRS containers during `oss-crs run`.

### Features

- **`oss-crs setup` command** - Configures the host system for cgroup-parent mode:
  - Sets Docker cgroup driver to `cgroupfs`
  - Enables cgroup v2 delegation (`cpuset`, `memory` controllers)
  - Creates user-owned `/sys/fs/cgroup/.../oss-crs/` directory
  - Supports `--check` (status only) and `--yes` (non-interactive) flags

- **Auto-detection of cgroup-parent availability** - Runtime detection at `build-target` and `run`:
  - If cgroup-parent is available: uses shared resource limits per CRS
  - If not: falls back to per-container `mem_limit` and `cpuset` limits

- **Resource limits for build containers** - Build containers now respect `cpuset` and `mem_limit` from `crs-compose.yaml`

### Other Changes

- Added `docs/setup.md` with usage and troubleshooting guide
- Fixed pytest as dev dependency instead of extra
- Added `dummy-crs` test fixture for resource limits testing

## Test Plan

- [ ] Run `oss-crs setup` and verify cgroup configuration
- [ ] Run `oss-crs setup --check` to verify status reporting
- [ ] Run `oss-crs build-target` and verify resource limits applied to build container
- [ ] Run `oss-crs run` with cgroup-parent mode and verify shared limits
- [ ] Run `pytest oss_crs/tests/integration/test_resource_limits.py`

Closes #18

### DIFF
--- a/oss_crs/src/templates/renderer.py
+++ b/oss_crs/src/templates/renderer.py
@@ -72,6 +72,7 @@ def render_build_target_docker_compose(
     build_out_dir: Path,
     build_id: str,
     sanitizer: str,
+    build_fetch_dir: Optional[Path] = None,
 ) -> str:
     """Render the docker-compose file for building a target.
 
@@ -82,6 +83,7 @@ def render_build_target_docker_compose(
         build_out_dir (Path): Output directory for the build.
         build_id (str): Build identifier.
         sanitizer (str): Sanitizer to use (e.g., "address", "undefined").
+        build_fetch_dir: Optional path to build-time fetch directory.
 
     Returns:
         str: Rendered docker-compose content as a string.
@@ -108,6 +110,7 @@ def render_build_target_docker_compose(
             "cpuset": crs.resource.cpuset if crs.resource else None,
             "memory": crs.resource.memory if crs.resource else None,
         },
+        "build_fetch_dir": str(build_fetch_dir) if build_fetch_dir else None,
     }
     return render_template(template_path, context)
 
--- a/docs/config/crs.md
+++ b/docs/config/crs.md
@@ -34,6 +34,7 @@ The root `CRSConfig` object contains the following fields:
 | `crs_run_phase` | `CRSRunPhase` | Yes | Configuration for the CRS run phase |
 | `supported_target` | `SupportedTarget` | Yes | Defines what targets this CRS supports |
 | `required_llms` | `list[string]` | No | Minimum required LLM model names (duplicates are automatically removed). This is for dependency validation, not a runtime allowlist. |
+| `required_inputs` | `list[string]` | No | Input channels this CRS requires to function. Valid names: `diff`, `pov`, `seed`, `bug-candidate`. When declared, `oss-crs run` validates that the corresponding CLI flags were provided before spawning containers. |
 
 ### Example
 
@@ -67,6 +68,9 @@ supported_target:
     - x86_64
 required_llms:
   - gpt-5
+required_inputs:
+  - diff
+  - bug-candidate
 ```
 
 ---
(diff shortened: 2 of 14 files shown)


oss_crs/src/templates/renderer.py
```diff
@@ -103,6 +103,10 @@ def render_build_target_docker_compose(
         "build_id": build_id,
         "crs_compose_env": crs.crs_compose_env.get_env(),
         "libCRS_path": str(LIBCRS_PATH),
+        "resource": {
+            "cpuset": crs.resource.cpuset if crs.resource else "0",
```
0xdkay: Is this ok as default values ? what about using all cores / all memory by default?

Is this actually used?

oss_crs/src/cgroup.py
```diff
@@ -0,0 +1,621 @@
...
+    # Set memory limit
+    (worker_path / "memory.max").write_text(str(total_memory_bytes))
+
+    # Enable controllers for sub-cgroups
+    (worker_path / "cgroup.subtree_control").write_text("+cpuset +memory")
+
+    return worker_path
+
+
+def create_crs_cgroup(
```
0xdkay, line 528: There is no subtree under `create_crs_cgroup` as in `create_worker_cgroup` is this intended?
Are we assuming that docker automatically handles that? 

```
  oss-crs/                                    ← subtree_control from setup
  └── worker-{run_id}/                        ← subtree_control: +cpuset +memory (code sets this)
      ├── crs-alpha/                          ← cpuset.cpus=0-3, memory.max=8G, NO subtree_control
      │   ├── <container-a1>/                 ← Docker auto-enables subtree_control on crs-alpha/
      │   └── <container-a2>/                 ← (second service in same CRS)
      ├── crs-beta/                           ← cpuset.cpus=4-7, memory.max=4G, NO subtree_control
      │   └── <container-b1>/                 ← Docker auto-enables subtree_control on crs-beta/
      └── crs-gamma/                          ← cpuset.cpus=4-7, memory.max=4G, NO subtree_control
          └── <container-g1>/                 ← Docker auto-enables subtree_control on crs-gamma/
```

Also, we are allowing overlapping cpusets, right?

oss_crs/src/cli/setup.py
```diff
@@ -0,0 +1,363 @@
...
+    ("oss-crs cgroup directory", check_directory, "directory"),
+    ("oss-crs controllers", check_controllers, "controllers"),
+]
+
+
+class SetupRunner:
+    """Runs the interactive setup process for cgroup-parent mode."""
+
+    def __init__(self, yes: bool = False):
+        self.console = Console(quiet=yes)
```
0xdkay: Should we make unified logging in utils.py and use that in overall code?

oss_crs/src/cgroup.py
azchin: Added subtree control at worker level.

oss_crs/src/templates/renderer.py
azchin: You're right, these are not good defaults. Changed to `None`, where limits are not imposed by default.

oss_crs/src/cli/setup.py
azchin: I moved many I/O components to `utils.py`. There was also code in `ui.py` that could use the refactored I/O helpers.

azchin:
> Also, we are delegating correctly handling stale cgroup or resource distribution to CRSBench, right?

How does CRSBench handle cgroups in the first place, since it doesn't have access to the OSS-CRS `compose.yaml` with CRS containers?

Do we have to disable cgroup management in OSS-CRS when CRSBench manages? I can add a flag for that, currently OSS-CRS will automatically set up cgroup parent.

Also, it turns out that my current implementation creates a new cgroup every time (even with same run-id) because it injects timestamp into the dir name. So there's never directory reuse, but will lead to empty cgroups left on disk over time (but apparently the kernel garbage collects them).

0xdkay:
How about this ownership model + interface?

  Ownership model:
  1. CRSBench owns the top-level `cgroup_parent` (trial root lifecycle).
  2. `oss-crs` owns only subtree creation under that parent for each run (`infra`, `crs-A/...`, `crs-B/...`).
  3. `oss-crs` maps containers/services to those subtree paths via `cgroup_parent`.
  4. `oss-crs` cleans only its owned run subtree (default behavior).
  5. CRSBench cleans the trial root.

  Interface:
  1. Expose `--cgroup-parent-root <path>` consistently on `prepare`, `build-target`, and `run`.
  2. Keep budget source-of-truth in CRSBench-generated `crs-compose.yaml` (no separate subtree-spec for now).
  3. `cpuset`/`mem_limit` per CRS are specified in `crs-compose.yaml`.

  Notes:
  1. We should keep `--cgroup-parent-root` even if `/proc/self/cgroup` is readable. Self-inspection is runtime observation/debugging, while explicit parent root gives deterministic placement and fail-fast validation.
  2. Please confirm if we want strict mode (no fallback) when `--cgroup-parent-root` is provided.

  Safety conditions:
  1. Subtree `cpuset` must be a subset of trial-root `cpuset`.
  2. If parent memory limit is set, `oss-crs` subtree/leaf memory limits must be `<=` parent limit.
  3. `oss-crs` should fail fast if parent path/delegation is invalid.
  4. Stale cleanup on both sides must be scoped strictly to owned prefixes only (never delete foreign/external cgroups).

PR #90, opened 2026-02-23
fix: stabilize target snapshot/run flow and builder base seeding for crsbench eval

0xdkay:
## Summary

This PR reorganizes and hardens the `fix/crsbench-eval` branch changes from `origin/main` with issue-focused commits.

### Included changes

1. `fix(target): stabilize injected Dockerfiles and snapshot creation`
- Stabilized generated Dockerfile handling in target image build.
- Added machine-level snapshot lock keyed by snapshot tag hash.
- Added deterministic intermediate snapshot tag and cleanup flow.
- Added atomic snapshot cache writes and target-scoped snapshot cache/out path keying.
- Updated snapshot unit tests.

2. `feat(run): capture compose logs and expose run artifacts`
- Added run-scoped log directory support.
- Added per-service log capture and per-CRS log grouping under run logs.
- Exposed run log paths via `oss-crs artifacts` output.
- Kept compose cleanup as hard-fail signal (teardown failure is not masked).
- Corrected compose log capture outputs (`docker-compose.stdout.log` / `.stderr.log`).
- Skips LLM validation when `llm_config` is not configured (null mode).

3. `fix(crs): include repository initialization failure details`
- `init_crs_repo` returns `TaskResult` instead of bool.
- Surfaces concrete repo-init failure detail in raised error.

4. `fix(builder): seed base build outputs from prebuilt artifacts`
- Builder sidecar now seeds `/builds/base/out` from `/OSS_CRS_BUILD_OUT_DIR/build`.
- Removes dependency on startup compile side effects for base build availability.

5. `docs(libcrs): document data exchange flow and run log artifacts`
- Added data exchange flow docs and run log artifact docs.
- Expanded `libCRS/README.md` command/usage description.

## Validation

- Unit tests:
  - `uv run pytest oss_crs/tests/unit`
  - Result: `37 passed`

## Notes

- There is an untracked local helper compose file in this branch workspace:
  - `example-compose.crs-claude-code.yaml`
  - It is not part of this PR unless explicitly added.

### DIFF
--- a/oss_crs/src/workdir.py
+++ b/oss_crs/src/workdir.py
@@ -66,10 +66,9 @@ class WorkDir:
         """Resolve a user-provided id to an existing directory name.
 
         Prefer exact match first, then try normalized form for backward
-        compatibility. IDs are treated as directory names, so path separators
-        are rejected.
+        compatibility.
         """
-        if not raw_id or "/" in raw_id or "\\" in raw_id:
+        if not raw_id:
             return None
         if (ids_dir / raw_id).is_dir():
             return raw_id
--- a/oss_crs/tests/unit/test_cli_artifacts.py
+++ b/oss_crs/tests/unit/test_cli_artifacts.py
@@ -27,12 +27,6 @@ def test_resolve_existing_id_returns_none_when_missing(tmp_path: Path) -> None:
     assert WorkDir._resolve_existing_id("does-not-exist", tmp_path) is None
 
 
-def test_resolve_existing_id_does_not_normalize_path_input(tmp_path: Path) -> None:
-    normalized = normalize_run_id("../escape")
-    _mkdir(tmp_path, normalized)
-    assert WorkDir._resolve_existing_id("../escape", tmp_path) is None
-
-
 def test_resolve_existing_id_prefers_exact_when_both_exist(tmp_path: Path) -> None:
     raw = "my-exact-id"
     normalized = normalize_run_id(raw)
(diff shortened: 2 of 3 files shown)


oss_crs/tests/unit/test_target_snapshot.py
azchin, line 1: This is really hard to understand. But as long as you think the following are ok, I think it's fine
- are the tests too tightly coupled with implementation (fragile on future impl change)?
- are they testing the problematic cases that you solved with this PR?

0xdkay: Yes, these tests are implementation-coupled intentionally.
- The fixes in this PR are internal orchestration behaviors (cache keying, atomic cache writes, deterministic intermediate tags), not a public API contract.
- The core tests directly target the failure modes fixed by this PR: stale cache reuse, non-atomic/colliding cache updates
- Concurrency lock contention is hard to validate in unit tests without real Docker/process parallelism, and I will check this with CRSBench.

oss_crs/src/workdir.py
```diff
@@ -60,6 +61,34 @@ def get_run_dir(self, run_id: str, sanitizer: str) -> Path:
...
 
+    @staticmethod
+    def _resolve_existing_id(raw_id: str, ids_dir: Path) -> str | None:
+        """Resolve a user-provided id to an existing directory name.
+
+        Prefer exact match first, then try normalized form for backward
+        compatibility. IDs are treated as directory names, so path separators
+        are rejected.
+        """
+        if not raw_id or "/" in raw_id or "\\" in raw_id:
```
azchin: Should we just move this invalid slash char logic into `normalize_run_id`? And get `normalize_run_id` to return `str | None`, these workdir helpers can just forward the `None` to caller.

0xdkay: make sense. As current `normalize_run_id` normalizes those characters to `-` we can remove that safely.

Complete via https://github.com/sslab-gatech/oss-crs/pull/90/commits/9b59c711bdea756a86a7c7668978764ea4492609

PR #91, opened 2026-02-23
feat(llm): add explicit internal/external litellm modes and align runtime semantics

### LINKED ISSUE #89: fix(llm): clarify required_llms semantics and external/null LiteLLM behavior
# LLM Mode Semantics: required_llms, null mode, and external LiteLLM support

## Summary

Current LLM handling has semantic drift and mode gaps:

1. `required_llms` is intended as a **minimum required dependency list**, but is currently also used as a strict key allowlist in internal LiteLLM key generation.
2. `llm_config: null` is intended to disable OSS-CRS-managed LiteLLM, but validation behavior has been inconsistent/confusing in practice.
3. First-class external LiteLLM mode is missing (previous functionality was not ported), forcing implicit env-var workarounds.

This causes runtime 401 failures (`key_model_access_denied`) when runtime model envs differ from `required_llms`, even when models are valid and available.

## Expected Behavior

- `required_llms` means: **minimum models that must exist/reachable**.
- `required_llms` does **not** restrict which model a CRS may use at runtime.
- `llm_config: null` means: no OSS-CRS LiteLLM checks, no internal sidecar/keygen injection.
- External LiteLLM should be supported explicitly in compose schema.

## Current Behavior

- Internal key-gen uses:
  - `models = required_llms`
- This effectively turns `required_llms` into an allowlist.
- Runtime model env (e.g. `ANTHROPIC_MODEL`) outside that list can fail with 401 despite model availability.

### 1) Clarify and enforce semantics

- Treat `required_llms` strictly as validation baseline.
- Do not use `required_llms` as strict key allowlist in internal mode.
- If allowlist is desired, add explicit optional `allowed_llms`.

### 2) Fix null mode contract

- If `llm_config` is null:
  - skip LLM validation
  - skip internal LiteLLM/postgres/keygen rendering
  - no auto-injected `OSS_CRS_LLM_API_URL` / `OSS_CRS_LLM_API_KEY`

### 3) Add first-class external mode

Suggested schema:

```yaml
llm_config:
  litellm_external:
    url: https://<external-litellm>
    key_env: LITELLM_API_KEY   # optional policy choice
```

Behavior:

- No internal LiteLLM sidecar/keygen services.
- Inject CRS runtime envs from external config.
- Preflight for external mode:
  - validate required fields (`url`, key if required)
  - `/models` reachability check
  - verify `required_llms ⊆ available_models`

### 4) Docs and migration guidance

- Define semantics of:
  - `required_llms`
  - `allowed_llms` (if introduced)
  - `llm_config: null`
  - `litellm_config` (internal)
  - `litellm_external` (external)
- Add decision table and examples.

## Acceptance Criteria

- [ ] `required_llms` no longer acts as implicit runtime allowlist.
- [ ] `llm_config: null` performs no LiteLLM checks and starts no LiteLLM infra.
- [ ] External LiteLLM mode works via compose config without workaround env hacks.
- [ ] External mode validates `/models` and required_llms presence.
- [ ] Documentation updated with explicit mode semantics and examples.
- [ ] Regression tests added for all three modes: null / internal / external.

## Context

Observed failure pattern:

- Key generated with models restricted to `required_llms`
- CRS runtime chose model not in that list
- LiteLLM returned:
  - `401 key_model_access_denied`

0xdkay:
## Summary

Implements issue #89 by unifying OSS-CRS LiteLLM configuration into explicit modes and fixing related runtime behavior.

### What changed

- Added explicit LiteLLM modes under `llm_config.litellm`:
  - `mode: internal | external`
  - `model_check` toggle (default `true`)
  - `internal.config_path` support (optional)
  - `external` one-of validation for `url|url_env` and `key|key_env`
- Added backward compatibility for legacy shape:
  - `llm_config.litellm_config` -> translated to internal mode
- Added bundled default internal LiteLLM config:
  - `oss_crs/defaults/litellm/default-models.yaml`
- External mode behavior:
  - injects external URL/key into CRS containers
  - skips internal LiteLLM sidecars/keygen
- Null mode behavior:
  - if `llm_config` is omitted/null, pre-run LLM validation is skipped
- Required model semantics:
  - `required_llms` treated as minimum dependency set (validation baseline), not key allowlist
- Internal key-gen fixes:
  - always generates keys even when `required_llms` is unset
  - keeps budget mapping as `llm_budget` -> LiteLLM `max_budget` (`null` when omitted)

### Docs/examples

- Updated compose/docs references for new schema and mode semantics
- Clarified internal key generation is independent of `required_llms`
- Clarified omitted `llm_budget` behavior (`max_budget: null`)
- Updated example compose files to `llm_config.litellm` format
- Updated `example/crs-claude-code/compose.yaml` default Anthropic model to `claude-opus-4-6`

## Validation

- Unit tests pass:
  - `uv run pytest oss_crs/tests/unit` (39 passed)
- Smoke-validated locally:
  - `crs-claude-code` + `sanity-mock-c-delta-01` (patching flow)
  - `atlantis-multilang-given_fuzzer` + `sanity-mock-c-delta-01` (PoV produced)

Closes #89

### DIFF
--- a/docs/design/libCRS.md
+++ b/docs/design/libCRS.md
@@ -67,6 +67,36 @@ libCRS relies on several environment variables injected by CRS Compose at contai
 
 libCRS uses the **strategy pattern** via the abstract `CRSUtils` base class. Currently, `LocalCRSUtils` implements all operations for local Docker Compose deployments. An `AzureCRSUtils` implementation is planned to support Azure-based deployments (e.g., using Azure Blob Storage for shared filesystems and Azure Container Instances for CRS execution). New deployment backends can be added by implementing the `CRSUtils` interface without changing the CLI or any CRS code.
 
+### Data Exchange Flow
+
+CRS containers never communicate directly. All inter-CRS data flows through a two-tier filesystem managed by the exchange sidecar:
+
+```
+  CRS-A Container                                              CRS-B Container
+ ┌──────────────────┐                                        ┌──────────────────┐
+ │                  │                                        │                  │
+ │  SubmitHelper    │                                        │  FetchHelper     │
+ │  (watchdog +     │                                        │  (poll every 5s) │
+ │   MD5 dedup +    │                                        │                  │
+ │   batch flush)   │                                        │  InfraClient     │
+ │       │          │                                        │    .fetch_new()  │
+ │       ▼          │                                        │       ▲          │
+ │  SUBMIT_DIR/     │       Exchange Sidecar                 │  FETCH_DIR/      │
+ │  ├─ povs/        │      (oss-crs-infra)                   │  ├─ povs/        │
+ │  ├─ seeds/       │    ┌──────────────────┐                │  ├─ seeds/       │
+ │  ├─ patches/     │───▶│  EXCHANGE_DIR/   │───▶            │  ├─ patches/     │
+ │  └─ bug-         │    │  (shared global) │                │  └─ bug-         │
+ │     candidates/  │    └──────────────────┘                │     candidates/  │
+ │                  │                                        │                  │
+ │  (per-CRS,       │                                        │  (per-CRS,       │
+ │   write-only)    │                                        │   read-only)     │
+ └──────────────────┘                                        └──────────────────┘
+```
+
+- **SUBMIT_DIR** — Per-CRS write area. `SubmitHelper` watches a local directory for new files, deduplicates by MD5 hash, batches them (100 files or 10 seconds), and copies to `SUBMIT_DIR/<type>/` with hash-based filenames.
+- **EXCHANGE_DIR** — Global shared area managed by the exchange sidecar. The sidecar copies files from each CRS's SUBMIT_DIR into EXCHANGE_DIR and distributes them to other CRSs' FETCH_DIRs.
+- **FETCH_DIR** — Per-CRS read-only area. `FetchHelper` polls every 5 seconds via `InfraClient`, deduplicates by filename (hash-based names from submit provide natural content dedup), and copies new files to the local directory.
+
 ### Data Types
 
 All submission and fetching operations work with one of the following data types:
--- a/docs/design/parallel.md
+++ b/docs/design/parallel.md
@@ -73,6 +73,7 @@ Run artifacts are stored at:
 {work_dir}/{sanitizer}/runs/{run-id}/crs/{crs-name}/{target}/SUBMIT_DIR/{harness}/patches/
 {work_dir}/{sanitizer}/runs/{run-id}/crs/{crs-name}/{target}/SHARED_DIR/{harness}/
 {work_dir}/{sanitizer}/runs/{run-id}/EXCHANGE_DIR/{target}/{harness}/      # shared across all CRSs
+{work_dir}/{sanitizer}/runs/{run-id}/logs/{target}/{harness}/              # compose/service logs
 ```
 
 ## Artifacts Command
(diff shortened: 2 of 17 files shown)


oss-crs-infra/litellm-key-gen/main.py
```diff
@@ -7,14 +7,14 @@
 LITELLM_API_URL = os.getenv("LITELLM_API_URL")
 
 
-def create_llm_key(key: str, budget: int, required_models: list[str]) -> str | None:
+def create_llm_key(key: str, budget: int, _required_models: list[str]) -> str | None:
```
azchin: Why this rename, is it no longer used? We can just remove the param if that's the case.

oss_crs/defaults/litellm/default-models.yaml
azchin, line 1: I was thinking using `OPENAI_API_KEY` etc. https://github.com/sslab-gatech/oss-crs/blob/v0.1.0/bug_finding/templates/config-litellm.yaml

oss_crs/src/llm.py
```diff
@@ -1,74 +1,214 @@
...
-                msg += f"  - {model}\n"
-            return TaskResult(success=False, error=msg.strip())
+
+        if not self.model_check_enabled:
+            return TaskResult(success=True)
+
+        if not required_llms:
+            return TaskResult(success=True)
+
+        if self.mode == "internal":
```
azchin, line 150: Deduplicate the shared code in `if self.mode == "internal"` and `if self.mode == "external"` blocks please

oss-crs-infra/litellm-key-gen/main.py
0xdkay: Removed via https://github.com/sslab-gatech/oss-crs/pull/91/commits/5fb39448345a4536e4e322092124c508e70fddfd

oss_crs/src/llm.py
0xdkay: Completed via https://github.com/sslab-gatech/oss-crs/pull/91/commits/5fb39448345a4536e4e322092124c508e70fddfd

oss_crs/defaults/litellm/default-models.yaml
0xdkay: One thing to consider:
- If we want to use this default-models.yaml as a fallback for the internal mode, we are assuming users have all those keys specified in the env vars?

azchin: No we don't need to assume users have all keys. I think it's more suitable for a "default" to use conventional provider env vars.
Do you know whether LiteLLM proxy will die if some keys are empty / invalid? Die from startup health checks or something

azchin: An alternative is we auto-populate the `default-models.yaml` based on what keys exist in environment, but then we're getting into heavy implicit logic.

0xdkay: Let me check if it dies with empty keys.

0xdkay: LiteLLM does not die on startup when provider key envs are missing.

azchin: Awesome, let's just keep the `default-models.yaml` static + rename the env vars to `OPENAI_API_KEY` etc

0xdkay: Yeah, lets use provider's env vars as-is in this case.

0xdkay: Completed via https://github.com/sslab-gatech/oss-crs/pull/91/commits/1a02dc3731341d148f6ca181e7dd56863dc74129

PR #95, opened 2026-02-24
feat: support directed build inputs for native CRS fuzzers

### LINKED ISSUE #92: Proposal: Support build-target --diff for native directed fuzzers (libCRS-aware)
# Support `build-target --diff` for native directed fuzzers (libCRS-aware)

## Problem
Directed fuzzers (AFLGo/SelectFuzz/Lyso/Titan-style) need diff/target context during **build-target** to generate directed artifacts (e.g., `BBtargets.txt`, `distance.cfg.txt`).

Today, diff is only available in `run` (`--diff` -> fetch dir), so directed CRSs must rebuild in run phase. This also risks cache misuse across different diffs.

## Proposal (minimal)
1. Add `--diff <file>` to `oss-crs build-target`.
2. Expose build-phase fetch mount/env so **libCRS can fetch diff in build containers** (same model as run):
   - `libCRS fetch diff <dir>` works in build-target containers.
3. Include `diff_sha256` in build/cache identity.
4. Persist build metadata with `diff_sha256`.
5. On `run --build-id ... --diff ...`, verify diff hash matches build metadata.

## Why libCRS
Use libCRS for build-phase diff access to keep CRS logic uniform:
- same fetch semantics in build and run
- no CRS-specific mount/path assumptions
- simpler CRS portability across environments

## Acceptance criteria
- `oss-crs build-target --diff ref.diff` works.
- Build containers can read diff via `libCRS fetch diff`.
- Different diffs do not collide in build cache.
- Run fails fast on build/diff hash mismatch.
- Non-directed flows unchanged when `--diff` is omitted.

## Out of scope
- Harness-specific build partitioning
- Cross-diff dedup/scheduling
- New run environments

0xdkay:
## Summary
- add `build-target --diff` support and pass directed inputs into build-phase `OSS_CRS_FETCH_DIR`
- split bug-candidate input flags for explicitness:
  - `--bug-candidate <file>`
  - `--bug-candidate-dir <dir>`
- include directed-input hash in build cache identity and persist build metadata:
  - `diff_sha256`
  - `bug_candidate_sha256`
  - combined `input_sha256`
- fail fast on `run --build-id` when provided directed inputs do not match build metadata
- keep build-phase input consumption aligned with libCRS fetch semantics (remove shortcut env vars)
- update docs for directed build inputs, fetch-dir behavior, and TODO to standardize bug-candidate format (linking `libCRS/libCRS/sarif.py`)

## Why Hashing
- directed build artifacts depend on input context (diff and/or bug-candidates)
- cache identity now uses a combined directed-input hash to prevent cross-input cache collisions
- persisted metadata enables deterministic mismatch checks before run

## Scalability Note
- directed-input hashing now streams file contents and hashes directory entries deterministically (sorted walk + relative path + file hash), avoiding large in-memory reads

## Notes
- bug-candidate files are currently passed through without runtime SARIF validation (intentional for now)

Closes #92

### DIFF
--- a/oss_crs/src/crs_compose.py
+++ b/oss_crs/src/crs_compose.py
@@ -433,7 +433,8 @@ class CRSCompose:
         if seed_dir is not None and not seed_dir.is_dir():
             print(f"Error: Seed directory does not exist: {seed_dir}")
             return False
-        if not self._validate_required_inputs(
+        if not self.__validate_before_run(
+            target,
             diff=diff,
             pov=pov,
             pov_dir=pov_dir,
@@ -441,8 +442,6 @@ class CRSCompose:
             bug_candidate=bug_candidate,
             bug_candidate_dir=bug_candidate_dir,
         ):
-            return False
-        if not self.__validate_before_run(target):
             return False
         target.init_repo()
 
@@ -538,11 +537,8 @@ class CRSCompose:
         seed_dir: Optional[Path] = None,
         bug_candidate: Optional[Path] = None,
         bug_candidate_dir: Optional[Path] = None,
-    ) -> bool:
-        """Validate that all CRS-declared required_inputs are provided.
-
-        Returns True if all requirements are satisfied (or no CRS declares any).
-        """
+    ) -> TaskResult:
+        """Validate that all CRS-declared required_inputs are provided."""
         provided: set[str] = set()
         if diff is not None:
             provided.add("diff")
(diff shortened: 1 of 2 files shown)


docs/crs-development-guide.md
```diff
@@ -756,6 +768,11 @@ The operator passes data via `oss-crs run`:
 - `--pov <file>` or `--pov-dir <dir>` — PoV files → `FETCH_DIR/povs/`
 - `--diff <file>` — Reference diff → `FETCH_DIR/diffs/ref.diff`
 - `--seed-dir <dir>` — Seed files → `FETCH_DIR/seeds/`
+- `--bug-candidate <file>` — Bug-candidate report file → `FETCH_DIR/bug-candidates/`
```
azchin, line 778: I see, so from the CRS point of view they only need to handle `FETCH_DIR/bug-candidates` directory (and not the single bug-candidate file case)?

What is the interaction between `--diff` and `--bug-candidates`? 
Are those 
1. mutually exclusive (either delta or sarif mode)
2. must be paired together (bug candidate only available in delta mode), or 
3. independent of each other?

Do we need to introduce a new mode for the CRSs to indicate whether they support `--bug-candidate`?
I.e. select modes support from {full, delta, sarif}?

0xdkay: `--diff` is explicitly for git diff like one, and `--bug-candidate` is for analysis reports like sarif; for now, we haven't decided the format yet. Said that, they are independent.

CRS developers can choose whether to use it or not, but for the registry, we can ask them to support `pov` and `seed` at least.  `--bug-candidate` is an experimental feature for now I think. (taking sarif reports as an input for directed fuzzers).

I think mode does not matter. `oss-crs` does not care about the mode, but users can use it as `full` or `delta` mode by providing the `--diff`.

azchin: I think mode does matter, because some CRSs can only run in certain modes. For example, can the directed fuzzers run without providing `--diff` or `--bug-candidate`?

Maybe not `mode`, and instead in `crs.yaml` we should ask CRS to specify what metadata they depend on?
```
required_metadata:
  - diff
  - sarif
```

0xdkay: That would be better. This case, bug finding / bug fixing type can be handled as well?
Bug finding --> at least support seed (as we are focusing on ensembling)
Bug fixing --> at least support pov (for patching)

We can make other optional I think.  diff, ... 
bug-candidate sounds a bit weird though ... sarif?

azchin: Hm there are two different semantics being discussed:
1. CRS requires the metadata to function
2. CRS provides functionality to process that metadata

Where (1) implies (2).

In the case of directed fuzzers, we may encounter a DGF that requires SARIF (bug location) to start up.

In the case of the bug-finding vs patching CRS, we need them to support seeds / povs as input for proper OSS-CRS integration, but a bug-finding CRS doesn't need the initial seeds to function properly (since initial seeds is an optional field).

0xdkay: Good point, I agree we should separate capability from requirement.
 In this PR, I’m keeping orchestration semantics as:

1. OSS-CRS provides standard input channels (seed / pov / bug-candidate / diff)
2. Whether an input is required is CRS/workflow-specific
  
So for directed fuzzers that require SARIF, the CRS should validate at startup and fail fast if missing (their job).

For generic bug-finding integration, I think if they support seeds, initial seeds should be supported as they are initially the same as the shared seeds during execution. Also, for ensembling, we want to make every registered bugfinding crs support seed sharing. So, this might be automatically handled at the starting point or during execution.

For patching CRS, povs can be optional, as they can support only diff or sarif only.

For now, lets keep bug-candidate and later if we have proper use case, then lets move on to like `sarif` or others.

azchin: Or have `required_metadata`, but only use it to specify requirements and not capabilities.

We can apply to other CLI options like `harness`. Create an error message through oss-crs to users saying "please provide --harness, --diff" etc. Catch the error earlier than after spawning CRS runtime.

0xdkay: Lets do like
```
  required_inputs:
    - diff
  required_llms:
    - gpt-4
```

0xdkay: Completed via https://github.com/sslab-gatech/oss-crs/pull/95/commits/5f0df1a643e85f59edb638d2c1ff76bf872e25f2

PR #105, opened 2026-03-04
feat: auto-clone target repo from main_repo in project.yaml

HQ1995:
## Summary
- When `--target-source-path` is not provided, automatically clone the repository specified by `main_repo` in `project.yaml`
- Added guard in `_compute_repo_key` for missing `main_repo`
- Fixed `Path("")` edge case in `_user_provided_repo` detection

## Problem
AIxCC targets (under `oss-fuzz/projects/aixcc/`) have `git clone` commented out in their Dockerfiles — they define `main_repo` in `project.yaml` and expect the source to be injected externally. Without `--target-source-path`, these targets fail during `docker build` with errors like:
```
/src/mock-c/mock.c: No such file or directory
```
(`sanity-mock-c-delta-01` as an example.)

Standard oss-fuzz projects are unaffected because their Dockerfiles contain `RUN git clone ...`, so the source is already present in the image regardless of whether `--target-source-path` is provided.

## Behavior after this change
1. User provides `--target-source-path` → use as-is (unchanged)
2. `project.yaml` has `main_repo`, no `--target-source-path` → auto-clone and inject
3. Neither → plain build without repo overlay (unchanged)

## Test plan
- [x] AIxCC target with `main_repo` (`sanity-mock-c-delta-01`): auto-clone + build succeeds
- [x] AIxCC target with explicit `--target-source-path`: uses provided path as-is
- [x] Standard oss-fuzz with `main_repo` (`libxml2`): auto-clone + build succeeds
- [x] Standard oss-fuzz without `main_repo` (`bzip2`): plain build succeeds

### DIFF
--- a/oss_crs/src/target.py
+++ b/oss_crs/src/target.py
@@ -90,14 +90,16 @@ class Target:
         self.work_dir = work_dir / "targets" / self.name
         self.work_dir.mkdir(parents=True, exist_ok=True)
 
-        self._user_provided_repo = repo_path is not None
-        if repo_path:
+        self._user_provided_repo = repo_path is not None and str(repo_path) != ""
+        if self._user_provided_repo:
             self.repo_path = repo_path
-        else:
+        elif self.config.main_repo:
             # Default repo path includes hash of repo URL to isolate parallel builds
             # of different repo sources (e.g., different forks or refs)
             repo_key = self._compute_repo_key()
             self.repo_path = work_dir / "targets" / f"{self.name}_{repo_key}" / "repo"
+        else:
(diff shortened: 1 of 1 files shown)


0xdkay:
The original intention of not including auto-fetch `main_repo` here, is that it is AIxCC-specific feature, which does not follow the official oss-fuzz workflow.

Accordingly, we ported all those AIxCC dataset in CRSBench, so you can use the dataset without cloning that.

If you think we still need that, we can add, but it will make a conflict in CRSBench as well, (which has `main_repo` in `project.yaml` as a reference).

PR #115, opened 2026-03-10
Patch ensemble infra & crs

occia:
## Summary

Support patch ensemble with a must-select strategy via `crs-patch-ensemble` integration and necessary infrastructure changes.

## Patch ensemble design

`crs-patch-ensemble` is designed as a CRS that consumes patches from other patch-producing CRSes and outputs the ensemble-selected patch. It validates each candidate (build + POV + test) and selects the best one to submit (must-select strategy, can be extended with more strategy when necessary). It can be composed with any combination of patch-producing CRSes.

As detailed in [crs-patch-ensemble README](https://github.com/Team-Atlanta/crs-patch-ensemble/blob/main/README.md), the ensemble CRS has two interface requirements:

1. **Patch listening** — The ensemble monitors the exchange directory (`OSS_CRS_FETCH_DIR`) for patches submitted by other CRSes. The existing **exchange sidecar** syncs each CRS's SUBMIT_DIR into the shared exchange directory, providing this capability.

2. **Ensemble timing** — The ensemble needs to know when all patch CRSes have exited (expected or unexpected) so it can make a final ensemble decision. A new **lifecycle sidecar** (`oss-crs-infra/lifecycle/`) monitors Docker container `die` events for all non-ensemble bug-fixing CRSes and writes a ready signal to `$exchange_dir/status/ready` when all have exited. This encapsulates all Docker-level logic so that `crs-patch-ensemble` itself has no Docker dependency — it only needs to know when all patches have arrived.

### Auto-attach for premature exit prevention
Docker Compose tears down all services when any attached service exits. If a patch CRS finishes before the ensemble, the ensemble would be killed prematurely. Solved by adding an `attach` field to CRS config and **auto-detecting** patch-ensemble presence — when a `bug-fixing-ensemble` CRS exists, the framework automatically sets `attach: false` on all other bug-fixing CRSes.

- Based on local testing, setting `attach: false` does not affect Docker Compose log storage.
- Only keeping the patch ensemble as attached naturally provides the desired behavior: killing the ensemble tears down all patch CRSes, while individual patch CRS exits do not affect the ensemble.

### CRS type semantics
The framework needs to distinguish ensemble CRSes from regular bug-fixing CRSes for proper template rendering. Added `bug-fixing-ensemble` as a new CRS type. `is_bug_fixing` returns `True` for both `bug-fixing` and `bug-fixing-ensemble`, so ensemble CRSes only need to declare `type: [bug-fixing-ensemble]`.

### Conditional sidecar deployment
Refined template conditions so sidecars are only deployed when needed:
- **Exchange sidecar**: multiple bug-finding CRSes (seed sharing) **or** patch-ensemble present (patch sharing)
- **Lifecycle sidecar**: only when a patch-ensemble CRS is present

### Tests
Added unit tests for:
- CRS type properties (`is_bug_fixing` implies for ensemble)
- Sidecar deployment conditions (exchange/lifecycle for different CRS compositions)
- Auto-attach logic (patch-ensemble triggers `attach: false` on other bug-fixing CRSes)

## End-to-end test result

Target: `atlanta-activemq-delta-01` (ActivemqOne) — crs-claude-code + crs-codex + crs-patch-ensemble

```
runs/1773111432ug/crs/
├── crs-claude-code/
│   └── SUBMIT_DIR/.../67b14203...57d.diff
├── crs-codex/
│   └── SUBMIT_DIR/.../7db9e3ad...b6f.diff
└── crs-patch-ensemble/
    ├── SUBMIT_DIR/.../67b14203...57d.diff       ← selected (= claude-code's patch)
    └── SHARED_DIR/ActivemqOne/
        ├── builds/
        │   ├── base/                             ← baseline build
        │   ├── (build-id-1)/                     ← claude-code patch (A) ✅ validated
        │   └── (build-id-2)/                     ← codex patch (B) ✅ validated
        ├── selector/run_1773111940/
        │   ├── prompt.txt                        ← full prompt with diffs + crash refs
        │   ├── selection.json                    ← {"selection": "A"}
        │   ├── claude_stdout.log
        │   └── claude_debug.log
        └── claude-logs/projects/.../<session>.jsonl  ← full Claude conversation transcript
```

**Selector decision**: Patch A (claude-code)

> Patch A is more semantically correct: it uses a comprehensive allowlist checked BEFORE `Class.forName()` to prevent untrusted class loading entirely, includes legitimate ActiveMQ exception packages (`javax.jms.*`, `jakarta.jms.*`, `org.apache.activemq.*`) preserving original functionality, and also blocks Error subclasses. Patch B only allows `java.*` which breaks legitimate server-side exception propagation for `javax.jms` and `org.apache.activemq` exceptions, and still calls `Class.forName()` before its safety check.

### Selector conversation transcript

```
L0:  [prompt] 2 candidate patches + crash log references
L4:  tool_use: Read("/work/base_crash_0/pov_stderr.log")
L7:  text: "The crash log confirms the vulnerability: Remote Code Execution
       via createThrowable() using network-controlled class names passed to
       Class.forName() + Constructor.newInstance()."
L8:  tool_use: Write("selection.json") → {"selection": "A", "reason": "..."}
L10: text: "Selection: A — Patch A checks the class name before Class.forName(),
       preventing untrusted classes from being loaded entirely..."
```

**Selector tools used**: `Read` (crash log) → `Write` (selection.json)

It didn't check the source code to make this decision (very obvious to pick by checking those 2 patches) while it has source code access.

### DIFF
--- a/oss-crs-infra/lifecycle/main.py
+++ b/oss-crs-infra/lifecycle/main.py
@@ -3,6 +3,11 @@
 Monitors Docker container events for a set of Compose services and writes
 a ready signal to the exchange directory once every watched service has
 stopped.  This lets an ensemble CRS know that no more artifacts will arrive.
+
+NOTE: This implementation relies heavily on the Docker Engine API (container
+events via /var/run/docker.sock).  To support other platforms (e.g. Kubernetes),
+the monitoring logic can be rewritten against their native APIs (e.g. Pod watch)
+and selected at runtime via an environment variable such as LIFECYCLE_BACKEND.
 """
 
 import json


oss_crs/src/templates/run-crs-compose.docker-compose.yaml.j2
```diff
@@ -157,10 +159,38 @@ services:
...
+{%- endfor %}
+{%- endif %}
+{%- endfor %}
+{%- if bug_fix_ensemble and exchange_dir and watch_services %}
+  oss-crs-lifecycle:
+    build:
+      context: {{ oss_crs_infra_root_path }}/lifecycle
+    attach: false
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock:ro
```
azchin, line 189: Originally I don't like this, having container manipulate host docker. Because it's not immediately translatable to remote deployment (e.g. k8s), and seems a bit hacky (e.g. what happens for users with weird Docker configurations, like TLS socket?).

But I think a similar approach can be done w/ k8s, where a pod can monitor others (given it's in same namespace or cluster).

If this is really the best approach (e.g. alternatives like heartbeat initiated by libcrs singleton pattern is not appropriate), then it's fine to keep. I just prefer something else.

oss_crs/src/config/crs.py
```diff
@@ -187,6 +188,14 @@ class CRSConfig(BaseModel):
     def is_builder(self) -> bool:
         return CRSType.BUILDER in self.type
 
+    @property
+    def is_bug_fixing(self) -> bool:
+        return CRSType.BUG_FIXING in self.type or CRSType.BUG_FIXING_ENSEMBLE in self.type
+
+    @property
+    def is_ensemble(self) -> bool:
```
azchin: Might get confusing if other ensemble stuff gets introduced.

oss_crs/src/crs_compose.py
```diff
@@ -50,8 +50,18 @@ def __init__(
         builder_count = sum(1 for crs in self.crs_list if crs.config.is_builder)
         if builder_count > 1:
             raise ValueError("At most one CRS entry with type 'builder' is allowed")
+
+        self._apply_patch_ensemble_auto_attach()
         self.deadline: Optional[float] = None
 
+    def _apply_patch_ensemble_auto_attach(self) -> None:
```
azchin: Is this overridable? If the user explicitly specified "attach: true" in their compose?

oss_crs/src/templates/run-crs-compose.docker-compose.yaml.j2
occia: I also don't like the mount of docker.sock (too rely on docker and require permission) but picked this instead of libCRS-based heartbeat due to:
- this additionally makes libCRS in charge of telemetry-like functionalities, i doubt about whether this fits in libCRS's position or not. Therefore, I tried to wrap all docker related logic into lifecycle sidecar and give patch ensemble crs an abstracted interface.

If adding heartbeat is well fitted into libCRS's scope, I'm happy to reimplement it inside libCRS.

oss_crs/src/config/crs.py
occia: thanks, solved.

oss_crs/src/crs_compose.py
occia: The `attach` parameter has been further removed from the user-facing YAML entirely.
When a patch-ensemble CRS is present, the framework now automatically sets attach: `false` on other bug-fixing CRS containers via the compose template, so no user configuration needed.

oss_crs/src/templates/run-crs-compose.docker-compose.yaml.j2
azchin: Yeah ok on second thought, implementing heartbeat reliably is also challenging since we're tackling a general distributed systems problem (maybe integrate sth like https://github.com/hashicorp/memberlist, seems quite overkill for libcrs).

Whereas reusing the existing infrastructure telemetry (docker, or in the future k8s) is more reasonable if we only need healthcheck.

occia: yes, implementing lifecycle sidecar using orchestration api is simpler. Let me refactor that sidecar logic a bit for easier switch to other orchestration frameworks in the future.

occia: > Let me refactor that sidecar logic a bit for easier switch to other orchestration frameworks in the future.

Emmm, this seems unnecessary since current logic is simple and tightly bound to docker api. Just added notes.

PR #118, opened 2026-03-11
fix: address post-#115 run and libCRS regressions

0xdkay:
## Summary

This PR addresses the full validated post-#115 regression set:

1. helper-sidecar teardown exits were being misclassified as run failures
2. non-ensemble runs could lose `OSS_CRS_FETCH_DIR`
3. `libCRS download-source repo` could resolve stale or wrong source paths

### 1. Classify helper teardown correctly
- use compose-event-aware shutdown classification
- avoid failing a run when helper sidecars exit with teardown-style codes after the primary service succeeds
- keep fallback behavior when compose event collection is unavailable
- keep compose-path handling consistent

### 2. Restore fetch dir for non-ensemble runs
- restore `OSS_CRS_FETCH_DIR` for normal single-CRS runs
- keep exchange-sidecar behavior scoped instead of globally re-enabling it

### 3. Resolve repo downloads from live source trees
- make `libCRS download-source repo` prefer the real live runtime source tree
- fall back sensibly when hints are stale or missing
- update related docs/changelog

## Validation

Targeted tests passed:
- `oss_crs/tests/unit/test_ui.py`
- `oss_crs/tests/unit/test_renderer_run_compose.py`
- `oss_crs/tests/unit/test_libcrs_download_source.py`

Smoke validation passed with real CRSes against `sanity-mock-c-delta-01`:
- `crs-codex`
- `crs-claude-code`
- `crs-gemini-cli`
- `crs-copilot-cli`

For the final smoke run:
- `oss-crs prepare` completed
- `oss-crs build-target` completed
- `oss-crs run --pov ... --timeout 300` completed
- all four CRSes submitted one patch on `fuzz_parse_buffer_section`

## Notes

- This is the same bug family discussed around #117, but with the full fix set rather than only the fetch/exchange symptom.

### DIFF
--- a/libCRS/libCRS/local.py
+++ b/libCRS/libCRS/local.py
@@ -79,6 +79,10 @@ class LocalCRSUtils(CRSUtils):
         """
         repo_hint = os.environ.get("OSS_CRS_REPO_PATH")
         runtime_root = Path(os.environ.get("SRC", "/src")).resolve()
+        build_out_dir = os.environ.get("OSS_CRS_BUILD_OUT_DIR") or None
+        build_out_src = (
+            Path(build_out_dir).resolve() / "src" if build_out_dir is not None else None
+        )
         if repo_hint:
             hinted_src = Path(repo_hint).resolve()
             if hinted_src.exists():
@@ -101,9 +105,13 @@ class LocalCRSUtils(CRSUtils):
                 continue
             return normalized_root
 
-        build_out_dir = os.environ.get("OSS_CRS_BUILD_OUT_DIR")
-        if build_out_dir:
-            build_out_src = Path(build_out_dir).resolve() / "src"
+        if build_out_src is not None:
+            if hinted_src is not None:
+                translated_hint = self._translate_repo_hint_to_build_output(
+                    hinted_src, runtime_root, build_out_src
+                )
+                if translated_hint is not None:
+                    return translated_hint
             normalized_build_out = self._normalize_repo_source_path(
                 build_out_src, build_out_src
             )
@@ -138,6 +146,29 @@ class LocalCRSUtils(CRSUtils):
             return candidate if candidate.exists() else None
 
         return None
+
+    def _translate_repo_hint_to_build_output(
+        self, hinted_src: Path, runtime_root: Path, build_out_src: Path | None
+    ) -> Path | None:
+        if build_out_src is None or not build_out_src.exists():
+            return None
+
+        relative_hint: Path | None = None
+        for source_root in (runtime_root, Path("/src").resolve()):
+            if hinted_src.is_relative_to(source_root):
+                relative_hint = hinted_src.relative_to(source_root)
+                break
+        if relative_hint is None:
+            return None
+
+        translated = build_out_src / relative_hint
+        if not translated.exists():
+            return None
+
+        normalized = self._normalize_repo_source_path(translated, build_out_src)
+        if normalized is not None:
+            return normalized
+        return translated
 
     def submit_build_output(self, src_path: str, dst_path: Path) -> None:
         src = Path(src_path)
--- a/oss_crs/tests/unit/test_libcrs_download_source.py
+++ b/oss_crs/tests/unit/test_libcrs_download_source.py
@@ -147,6 +147,34 @@ def test_download_source_repo_falls_back_to_build_output_src_when_hint_missing(
     assert calls["dst"] == dst.resolve()
 
 
+def test_download_source_repo_translates_stale_src_hint_to_build_output(
+    monkeypatch, tmp_path: Path
+) -> None:
+    build_out_dir = tmp_path / "build-out"
+    repo = build_out_dir / "src" / "mock-c"
+    repo.mkdir(parents=True)
+    (repo / "mock.c").write_text("ok")
+    dst = tmp_path / "dst"
+
+    monkeypatch.setenv("SRC", str(tmp_path / "missing-src-root"))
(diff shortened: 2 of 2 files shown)


libCRS/libCRS/local.py
```diff
@@ -64,6 +63,85 @@ def download_source(self, source_type: SourceType, dst_path: Path) -> None:
...
+        source_root = source_root.resolve()
+        if not candidate.exists() or not source_root.exists():
+            return None
+
+        try:
+            candidate.relative_to(source_root)
+        except ValueError:
+            return candidate if candidate.exists() else None
+
+        repo_dir = source_root / "repo"
```
azchin: Where is `/src/repo` a standard? I'm not familiar with many projects doing this. CRSBench?

libCRS/libCRS/local.py
```diff
@@ -64,6 +63,85 @@ def download_source(self, source_type: SourceType, dst_path: Path) -> None:
...
+
+    def _normalize_repo_source_path(
+        self, candidate: Path, source_root: Path
+    ) -> Path | None:
+        candidate = candidate.resolve()
+        source_root = source_root.resolve()
+        if not candidate.exists() or not source_root.exists():
+            return None
+
+        try:
```
azchin:
```
if not candidate.is_relative_to(source_root)
```

oss_crs/src/ui.py
```diff
@@ -800,6 +836,216 @@ def _check_failed_containers(
...
+            compose_data = yaml.safe_load(docker_compose_path.read_text()) or {}
+        except Exception:
+            return set()
+
+        services = compose_data.get("services", {})
+        helper_services = set()
+        for service_name, service_config in services.items():
+            if not isinstance(service_config, dict):
+                continue
+            if service_name.startswith("oss-crs-"):
```
azchin: Magic string, this is fine to keep. But let's document for developers that if we're adding new infra containers, they must start with "oss-crs-".

azchin (commented):
Nitpicks

oss_crs/src/ui.py
0xdkay: Completed via https://github.com/sslab-gatech/oss-crs/pull/118/commits/be1be00ba88955ea2d27e87522185ce275ac3765

libCRS/libCRS/local.py
0xdkay: Completed via https://github.com/sslab-gatech/oss-crs/pull/118/commits/5f2f6c7eae8de3e6c5963fb41fd21409e5c3e546

libCRS/libCRS/local.py
0xdkay: Completed via https://github.com/sslab-gatech/oss-crs/pull/118/commits/5f344deed54f3509e1ff8abe574eb75d54e8e847

0xdkay:
For b2a2988

  Current source flow is split across two phases:

  - In `build-target`, the CRS builder preserves the broad source workspace by submitting `$SRC` as the `src` build output. That ends up stored under `OSS_CRS_BUILD_OUT_DIR/src`.
  - In `run`, `download-source repo` does not read from a dedicated repo artifact. Instead, it uses `OSS_CRS_REPO_PATH` as a hint for where the project source is expected to be. Today that hint usually comes from the target’s effective Dockerfile `WORKDIR`.

  Because of that split, the hint can be stale at run time. For example, `OSS_CRS_REPO_PATH` may say `/src/mock-c`, while the actual preserved source available to the CRS is under `OSS_CRS_BUILD_OUT_DIR/src/mock-c`.

  This commit fixes that mismatch: when the hinted repo path no longer exists in the run container, `libCRS` now maps that hinted subpath onto the preserved source tree under `OSS_CRS_BUILD_OUT_DIR/src/...` instead of failing on the stale original path.

  This is a compatibility fix for the current model. A cleaner long-term design would expose explicit source artifacts directly from `oss-crs` across both `build-target` and `run`, so `download-source repo` would not need to reconcile a run-phase hint with build-phase preserved source.

libCRS/libCRS/local.py
```diff
@@ -64,6 +63,113 @@ def download_source(self, source_type: SourceType, dst_path: Path) -> None:
...
+            return None
+
+        if not candidate.is_relative_to(source_root):
+            return candidate if candidate.exists() else None
+
+        git_dir = next(source_root.rglob(".git"), None)
+        if git_dir is not None:
+            return source_root
+
+        if candidate != source_root:
```
azchin, line 145: So this condition is: if candidate existis, is subdir of `/src`, and src has no `.git` subdir, then return candidate.
E.g. `/src/nested/project` with no `.git` under `/src` --> `/src/nested/project` instead of `/src`.

What case does this accommodate? CRSBench with `pkgs/` subdir?

0xdkay: Yes, your read is correct.

  And yes, a `pkgs/`-style sibling directory is exactly the kind of case this is handling (for example, the curl-XX).

  Example:
  - `/src/curl` is the real project/workdir hint
  - sibling trees under `/src` also exist, such as `curl_fuzzer`, `zlib`, `openssl`, `nghttp2`, or `pkgs`
  - `/src` itself has no single `.git` marking the whole workspace as one repo

  In that case, giving the CRS all of `/src` is too broad, so `libCRS` keeps the narrower hinted path (for example `/src/curl`).

  This is a compatibility choice for the current CRSBench / OSS-Fuzz source model, not a perfect long-term source contract. A cleaner future design would likely require centralized, oss-crs-owned source management instead of inferring source shape from workspace layout and `.git` signals. (we may need to support non-git repos?)

PR #119, opened 2026-03-11
fix: always set exchange_dir for user-provided seeds, povs, and diffs

acorn421:
## Summary
Always set `exchange_dir` regardless of ensemble mode, since all CRS containers need
the FETCH_DIR mount to receive user-provided seeds, POVs, diffs, and bug-candidates.

## Description
Previously, `exchange_dir` was only set when running in ensemble mode
(`bug_finding_ensemble` or `bug_fix_ensemble`). This meant that in single-CRS runs,
the exchange directory path was empty, preventing users from providing seeds, POVs,
diffs, and bug-candidates via the FETCH_DIR mount.

This was discovered while running a bug-fixing CRS on CRSBench — ground truth POVs
were not being delivered to the container. Additionally, for corpus bootstrapping in
bug-finding CRS scenarios, the fetch directory must also be available by default.

Since always providing the exchange directory has no significant side effects, this
fix removes the conditional and always computes the exchange directory path.

### DIFF
--- a/oss_crs/src/templates/renderer.py
+++ b/oss_crs/src/templates/renderer.py
@@ -214,12 +214,9 @@ def render_run_crs_compose_docker_compose(
     )
     bug_finding_ensemble = bug_finding_crs_count > 1
     bug_fix_ensemble = any(crs.config.is_bug_fixing_ensemble for crs in crs_compose.crs_list)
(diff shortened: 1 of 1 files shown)


0xdkay:
This is fixed in #118 , introduced by #115

PR #137, opened 2026-03-19
feat: add headless plain-log fallback for non-interactive runs

fuyu0425:
## Summary
- disable Rich Live rendering when the console is non-interactive
- print task status, command metadata, and streamed command output as plain logs in headless mode
- add a regression test covering streamed output on a non-interactive console

## Verification
- uv run pytest -q oss_crs/tests/unit/test_ui.py
- uv run oss-crs build-target --compose-file ./example/crs-libfuzzer/compose.yaml --fuzz-proj-path ~/oss-fuzz/projects/libxml2


azchin:
Tested with tee

PR #168, opened 2026-04-15
fix: use Docker Compose secrets for LLM keys (#160)

### LINKED ISSUE #160: fix: use secrets in Docker compose instead of environment variables
https://docs.docker.com/compose/how-tos/use-secrets/

Currently we use env vars for our LLM keys. 

Such as the master key in https://github.com/ossf/oss-crs/blob/ceb67e0813f5d336165fd7ea394572deea59da61/oss_crs/src/templates/run-crs-compose.docker-compose.yaml.j2#L98 

And the per-CRS keys (change how they are passed from provisioner to CRS).

Let's use secrets instead for best practice and security.

tusharshah21:
fix: use Docker Compose secrets for LLM keys

## Summary

Replace environment variable delivery of `LITELLM_MASTER_KEY`, `POSTGRES_PASSWORD`, and per-CRS `OSS_CRS_LLM_API_KEY` with Docker Compose file-based secrets per the Docker Compose secrets best practice
(https://docs.docker.com/compose/how-tos/use-secrets/).

Previously, sensitive LLM keys were injected as plaintext environment variables into containers. They are now delivered as Docker secrets mounted at `/run/secrets/`, preventing accidental exposure via `docker inspect`, logs, or child process env inheritance.

## User Impact

- [x] User-facing change

CRS authors who read `OSS_CRS_LLM_API_KEY` directly from the environment
must update to read from the file path provided by `OSS_CRS_LLM_API_KEY_FILE`:
Closes #160 
```python
# Before
api_key = os.environ["OSS_CRS_LLM_API_KEY"]

# After
with open(os.environ["OSS_CRS_LLM_API_KEY_FILE"]) as f:
    api_key = f.read().strip()

### DIFF
--- a/oss-crs-infra/litellm-key-gen/main.py
+++ b/oss-crs-infra/litellm-key-gen/main.py
@@ -3,7 +3,17 @@ import os
 import yaml
 import requests
 
-LITELLM_MASTER_KEY = os.getenv("LITELLM_MASTER_KEY")
+
+def _read_secret(path: str) -> str | None:
+    """Read a Docker secret file."""
+    try:
+        with open(path) as f:
+            return f.read().strip()
+    except FileNotFoundError:
+        return None
+
+
+LITELLM_MASTER_KEY = _read_secret("/run/secrets/litellm_master_key")
 LITELLM_API_URL = os.getenv("LITELLM_API_URL")
 
 
--- a/oss-crs-infra/litellm-key-gen/main.py
+++ b/oss-crs-infra/litellm-key-gen/main.py
@@ -4,13 +4,10 @@ import yaml
 import requests
 
 
-def _read_secret(path: str) -> str | None:
+def _read_secret(path: str) -> str:
     """Read a Docker secret file."""
-    try:
-        with open(path) as f:
-            return f.read().strip()
-    except FileNotFoundError:
-        return None
+    with open(path) as f:
+        return f.read().strip()
 
 
 LITELLM_MASTER_KEY = _read_secret("/run/secrets/litellm_master_key")
--- a/oss_crs/src/templates/run-crs-compose.docker-compose.yaml.j2
+++ b/oss_crs/src/templates/run-crs-compose.docker-compose.yaml.j2
@@ -31,14 +31,22 @@ services:
       args:
         - target_base_image={{target.get_docker_image_name()}}
         - crs_version={{ crs.config.version }}
+{%- if snapshot_image_tag and (module_config.run_snapshot or crs.config.is_builder) %}
+        - snapshot_image={{ snapshot_image_tag }}
+{%- endif %}
     environment:
 {%- for env_key, env_value in module_envs[crs.name + "_" + module_name].items() %}
       - {{ env_key }}={{ env_value }}
 {%- endfor %}
-{%- if llm_context is defined and llm_context.mode == "internal" %}
+{%- if llm_context is defined and llm_context.mode == "internal" and not crs.config.is_builder %}
     depends_on:
       oss-crs-litellm-key-gen:
         condition: service_completed_successfully
+{%- endif %}
+{%- if llm_context is defined and crs.name in llm_context.secret_files %}
+    secrets:
+      - source: oss_crs_llm_api_key_{{ crs.name }}
+        target: oss_crs_llm_api_key
 {%- endif %}
 {%- if cgroup_parents and crs.name in cgroup_parents %}
     # Using cgroup-parent mode: resources managed by parent cgroup
@@ -84,12 +92,15 @@ services:
     depends_on:
       oss-crs-postgres:
         condition: service_healthy
+{%- if llm_context.litellm_env %}
     environment:
-      - LITELLM_MASTER_KEY={{ llm_context.litellm_master_key }}
-      - DATABASE_URL=postgresql://{{ postgres_user }}:{{ llm_context.postgres_password }}@{{ postgres_host }}:{{ postgres_port }}/{{ postgres_user }}
     {%- for key, value in llm_context.litellm_env.items() %}
       - {{ key }}={{ value }}
     {%- endfor %}
+{%- endif %}
+    secrets:
+      - litellm_master_key
+      - postgres_password
     volumes:
       - {{ llm_context.litellm_config_path }}:/app/config.yaml:ro
     networks:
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by azchin** [comment]
Thanks for the PR! I'll take a proper look over the weekend, in the meanwhile could you sign your commits (https://github.com/ossf/oss-crs/blob/main/CONTRIBUTING.md#developer-certificate-of-origin-dco) and force push?

We also updated main with a major internal refactor, hence the conflicts.

**Comment by tusharshah21** [comment]
Done mate, covered the conflicts too

**Comment by tusharshah21** [comment]
any updates?

File: oss-crs-infra/litellm-key-gen/main.py [hunk 1]
```diff
@@ -3,7 +3,17 @@
...
+def _read_secret(path: str) -> str | None:
+    """Read a Docker secret file."""
+    try:
+        with open(path) as f:
+            return f.read().strip()
+    except FileNotFoundError:
+        return None
+
+
+LITELLM_MASTER_KEY = _read_secret("/run/secrets/litellm_master_key")
```
azchin, line 14: We can let the `FileNotFoundError` terminate the process, easier to catch deployment bugs instead of later failure due to `LITELLM_MASTER_KEY = None`

oss_crs/src/templates/run-crs-compose.docker-compose.yaml.j2
```diff
@@ -31,15 +31,23 @@ services:
       args:
         - target_base_image={{target.get_docker_image_name()}}
         - crs_version={{ crs.config.version }}
+{%- if snapshot_image_tag and (module_config.run_snapshot or crs.config.is_builder) %}
```
azchin: I think this is from an improperly resolved conflict, shouldn't be in this PR's changes.

oss_crs/src/templates/run-crs-compose.docker-compose.yaml.j2
```diff
@@ -31,15 +31,23 @@ services:
...
+        - snapshot_image={{ snapshot_image_tag }}
+{%- endif %}
     environment:
 {%- for env_key, env_value in module_envs[crs.name + "_" + module_name].items() %}
       - {{ env_key }}={{ env_value }}
 {%- endfor %}
-{%- if llm_context is defined and llm_context.mode == "internal" %}
+{%- if llm_context is defined and llm_context.mode == "internal" and not crs.config.is_builder %}
```
azchin: Same here regarding conflicts.

oss_crs/src/templates/run-crs-compose.docker-compose.yaml.j2
```diff
@@ -160,7 +179,7 @@ services:
   ###########################################
   # OSS-CRS-LIFECYCLE SIDECAR
   ###########################################
-{#- Collect non-builder, non-ensemble bug-fixing CRS services #}
+{#- Collect non-ensemble bug-fixing CRS services #}
```
azchin: Improperly resolved merge conflict?

oss_crs/src/templates/run-crs-compose.docker-compose.yaml.j2
```diff
@@ -84,20 +92,28 @@ services:
     depends_on:
       oss-crs-postgres:
         condition: service_healthy
+{%- if llm_context.litellm_env %}
     environment:
-      - LITELLM_MASTER_KEY={{ llm_context.litellm_master_key }}
-      - DATABASE_URL=postgresql://{{ postgres_user }}:{{ llm_context.postgres_password }}@{{ postgres_host }}:{{ postgres_port }}/{{ postgres_user }}
     {%- for key, value in llm_context.litellm_env.items() %}
```
azchin: I think these need to be converted to secrets as well. This is where the user can provide e.g. `OPENAI_API_KEY`, and their LiteLLM-defined config picks it up via `os.environ/OPENAI_API_KEY`

What we should do here is create a wrapper script that somehow enumerates all the secret files we need (maybe convey it through a JSON or some other KVP encoding) and set the environment vars for the main LiteLLM process. Similar to how you already do `export ... && exec litellm` but for an arbitrary number of {environment variables : secret file}.

Unfortunately LiteLLM only supports env vars for keys in their proxy, they don't support secret files. https://docs.litellm.ai/docs/proxy/configs 

The following is an example, with `EXTERNAL_LITELLM_API_KEY`.
https://github.com/ossf/oss-crs/blob/930a3092fe2f405391343a82ef075bad7aaa016d/example/crs-codex/litellm-config.yaml#L12

azchin (changes requested):
The main thing missing is converting LiteLLM config environment variables into secrets.

The flow is OSS-CRS picks up environment variables from the host according to `litellm-config.yaml`, and exposes them as env vars for the guest LiteLLM proxy Docker containers. See below for details.

azchin:
Also since `OSS_CRS_LLM_API_KEY` --> `OSS_CRS_LLM_API_KEY_FILE` is a breaking change, I will need to update all of the existing CRSs in the registry.

tusharshah21:
my bad for the delays reporting back in a while

oss-crs-infra/litellm-key-gen/main.py
tusharshah21: fail-fast on missing secret
 Implemented. I removed the FileNotFoundError fallback, so missing now terminates immediately instead of continuing with None.

oss_crs/src/templates/run-crs-compose.docker-compose.yaml.j2
tusharshah21: snapshot_image block appears in this PR. That was conflict residue ;) I removed it from this PR and restored the upstream template behavior.

oss_crs/src/templates/run-crs-compose.docker-compose.yaml.j2
tusharshah21: I reverted this to the upstream

oss_crs/src/templates/run-crs-compose.docker-compose.yaml.j2
tusharshah21: I cleaned this up

tusharshah21:
check now  mate!

azchin:
Looks great! Can you resolve the CI errors? `uv run verify` should tell you what's wrong locally.

Please also apply this diff, I updated `crs-codex.yaml` on main to match your updates.

```diff
diff --git a/registry/crs-codex.yaml b/registry/crs-codex.yaml
index 7c542ea..622174b 100644
--- a/registry/crs-codex.yaml
+++ b/registry/crs-codex.yaml
@@ -3,4 +3,4 @@ type:
   - bug-fixing
 source:
   url: https://github.com/Team-Atlanta/crs-codex.git
-  ref: refactor/sidecar
+  ref: main
```

azchin:
Thanks for your contribution!

PR #169, opened 2026-04-15
fix(renderer): resolve oss-crs-infra: prefix in build-target context

acorn421:
## Summary

`render_build_target_docker_compose` builds the rendered docker-compose
`build.dockerfile` value by joining `build_config.dockerfile` directly under
`crs.crs_path`, skipping the `_resolve_module_dockerfile` helper that other
render paths already use. As a result, framework-infra references such as
`oss-crs-infra:default-builder` get rendered as a literal path
`<crs_path>/oss-crs-infra:default-builder`, and `docker compose` then aborts
with `failed to read dockerfile: open oss-crs-infra:default-builder: no such
file or directory`.

## Reproduction

A CRS that declares an `inc-builder` step with a framework reference, e.g.

```yaml
target_build_phase:
  - name: inc-builder
    snapshot: true
    dockerfile: oss-crs-infra:default-builder
```

triggers the bug when the build-target compose file is rendered for that
build step. Concretely, this is hit by the `crs-claude-code` registry entry
on its `inc-builder` build, and surfaces in CRSBench's smoke tests as:

```
[smoke:bugfixing:worker] failed to solve: failed to read dockerfile:
open oss-crs-infra:default-builder: no such file or directory
[smoke:bugfixing:worker] FAIL Prepare docker images defined in docker compose file
```

## Root cause

`oss_crs/src/templates/renderer.py` already exposes
`_resolve_module_dockerfile(crs_path, dockerfile)` (lines 36–47), which
resolves `oss-crs-infra:<module>` to
`OSS_CRS_ROOT_PATH/oss-crs-infra/<module>/Dockerfile` and falls back to
`crs_path / dockerfile` for plain paths. The `crs_run_phase` rendering path
routes its values through this helper. The build-target context builder
(line 129) was the only call site that bypassed it, so the prefix mechanism
silently failed for any infra reference declared in `target_build_phase`.

This was latent until a CRS started referencing `oss-crs-infra:` from
`target_build_phase`; before that, only `crs_run_phase` exercised the
prefix and the resolver was always applied.

## Fix

Route the build-target context's `builder_dockerfile` through the same
helper:

```python
"builder_dockerfile": _resolve_module_dockerfile(
    crs.crs_path, build_config.dockerfile
),
```

No behavioral change for plain dockerfile paths — `_resolve_module_dockerfile`
falls back to `crs_path / dockerfile` for any value that does not start with
`OSS_CRS_INFRA_PREFIX` (`"oss-crs-infra:"`).

## Tests

Added `oss_crs/tests/unit/test_renderer_build_target.py` with four cases:

- `test_infra_prefix_resolves_to_framework_dockerfile_path` — `oss-crs-infra:default-builder`
  resolves to `OSS_CRS_ROOT_PATH/oss-crs-infra/default-builder/Dockerfile`.
- `test_plain_dockerfile_resolves_relative_to_crs_path` — plain paths still
  resolve to `crs.crs_path / dockerfile`.
- `test_infra_prefix_does_not_leak_colon_into_path` (parametrized) — the
  literal `oss-crs-infra:` token never appears in the rendered dockerfile
  path, and the path ends with `/Dockerfile`.

All four pass; existing `test_renderer_run_compose`, `test_template_rendering`,
and `test_templates` continue to pass unchanged.

## Test plan

- [x] `uv run pytest oss_crs/tests/unit/test_renderer_build_target.py -v`
- [x] `uv run pytest oss_crs/tests/unit/test_renderer_run_compose.py oss_crs/tests/unit/test_template_rendering.py oss_crs/tests/unit/test_templates.py`
- [x] `uv run ruff check oss_crs/tests/unit/test_renderer_build_target.py`
- [x] `uv run ruff format --check oss_crs/tests/unit/test_renderer_build_target.py oss_crs/src/templates/renderer.py`
- [ ] CRSBench-sidecar smoke (`scripts/ci-tests/run-local.sh smoke-bugfixing`) once submodule pointer is bumped to this commit

### DIFF
--- a/oss_crs/src/templates/renderer.py
+++ b/oss_crs/src/templates/renderer.py
@@ -126,7 +126,9 @@ def render_build_target_docker_compose(
         "crs": {
             "name": crs.name,
             "path": str(crs.crs_path),
-            "builder_dockerfile": str(crs.crs_path / build_config.dockerfile),
+            "builder_dockerfile": _resolve_module_dockerfile(
+                crs.crs_path, build_config.dockerfile
+            ),
             "version": crs.config.version,
         },
         "effective_env": env_plan.effective_env,
--- a/oss_crs/tests/unit/test_renderer_build_target.py
+++ b/oss_crs/tests/unit/test_renderer_build_target.py
@@ -1,0 +1,112 @@
+"""Unit tests for ``render_build_target_docker_compose``.
+
+Focus: the ``crs.builder_dockerfile`` value rendered into the build-target
+docker-compose file must be resolved through ``_resolve_module_dockerfile``
+so that ``oss-crs-infra:<module>`` framework references expand to the infra
+Dockerfile path instead of being treated as a path under the CRS repo.
+"""
+
+from pathlib import Path
+from types import SimpleNamespace
+
+import pytest
+import yaml
+
+from oss_crs.src.templates.renderer import (
+    OSS_CRS_ROOT_PATH,
+    render_build_target_docker_compose,
+)
+
+
+def _patch_env(monkeypatch) -> None:
+    monkeypatch.setattr(
+        "oss_crs.src.templates.renderer.build_target_builder_env",
+        lambda **_kwargs: SimpleNamespace(effective_env={"EXAMPLE": "1"}, warnings=[]),
+    )
+
+
+def _make_crs(tmp_path: Path) -> SimpleNamespace:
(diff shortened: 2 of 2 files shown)


azchin:
I have questions... the reproduction `crs.yaml` is from the main branch of `crs-claude-code` which doesn't have the sidecar branch changes merged yet.

PR #238, opened 2026-05-28
feat(config): validate required_envs before run

### LINKED ISSUE #143: feat(config): add required_envs field to CRS config for mandatory environment variable validation
## Summary

CRS implementations that depend on host environment variables (e.g., API tokens) have no way to declare them at the framework level. Each CRS must implement its own ad-hoc validation logic, leading to late runtime failures.

## Description

Currently, CRSConfig in crs.yaml supports required_llms and required_inputs to declare dependencies that oss-crs validates before execution. However, there is no equivalent mechanism for required environment variables.

Multiple CRS implementations rely on host environment variables passed through via additional_env in compose YAML (e.g., `COPILOT_GITHUB_TOKEN: ${COPILOT_GITHUB_TOKEN}` in crs-copilot-cli, or LLM provider API keys like `OPENAI_API_KEY`, `ANTHROPIC_API_KEY` for LiteLLM-based CRS). Without a framework-level declaration, there is no preflight validation that the referenced variables actually exist on the host — failures only surface at runtime.

## Current Status

- required_llms: validates LLM model availability via validate_required_llms() — exists
- required_inputs: validates input artifact dependencies — exists
- required_envs: validates host environment variables — missing

## Proposed Solution

- Add a required_envs field to CRSConfig (similar to required_llms) that lists environment variable names the CRS requires
- Add a validate_required_envs() step in the compose preflight phase (crs_compose.py) that checks whether all declared env vars are present in the host environment or resolved via additional_env

bmendonca3:
## Summary
- add `required_envs` to CRSConfig with environment variable name validation
- validate CRS-required env vars during run preflight against host env and compose `additional_env`
- handle unresolved `additional_env` placeholders such as `${API_KEY}` as missing when the host env is unset
- add unit tests plus docs and changelog updates

Fixes #143.

## Testing
- `uv run verify`

### DIFF
--- a/oss_crs/src/crs_compose.py
+++ b/oss_crs/src/crs_compose.py
@@ -37,6 +37,23 @@ import docker.errors
 import requests.exceptions
 
 
+_ENV_INTERPOLATION_RE = re.compile(
+    r"(?<!\$)\$(?:\{([A-Za-z_][A-Za-z0-9_]*)(?:(:?[-?+])[^}]*)?\}|([A-Za-z_][A-Za-z0-9_]*))"
+)
+
+
+def _additional_env_value_is_resolved(value: object, host_envs: set[str]) -> bool:
+    """Return whether a compose additional_env value can be resolved now."""
+    for match in _ENV_INTERPOLATION_RE.finditer(str(value)):
+        env_name = match.group(1) or match.group(3)
+        operator = match.group(2)
+        if operator in ("-", ":-"):
+            continue
+        if env_name not in host_envs:
+            return False
+    return True
+
+
 class CRSCompose:
     @classmethod
     def from_yaml_file(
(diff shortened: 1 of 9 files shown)


oss_crs/src/crs_compose.py
```diff
@@ -37,6 +37,23 @@
...
+    for match in _ENV_INTERPOLATION_RE.finditer(str(value)):
+        env_name = match.group(1) or match.group(3)
+        operator = match.group(2)
+        if operator in ("-", ":-"):
+            continue
+        if env_name not in host_envs:
+            return False
+    return True
+
+
```
azchin: Is putting this in `env_schema.py` or `env_policy.py` appropriate?

I like this helper, can you also use it to give a *warning* when something in `additional_env` does not resolve to host but is not in `required_env`?

My use case is https://github.com/ossf/oss-crs/blob/11409f212554af208566ce64650376d8b324f27d/example/crs-bug-finding-claude-code/compose-oauth.yaml#L29
where the CRS does not require that env, but it will work if that env is provided.
But if I forget to set it, I do not want to wait for the CRS itself to error out before realizing.

azchin (commented):
I end-to-end tested with `crs-bug-finding-copilot-cli`, `required_envs` looks good.

Please see my suggestion and feature request commented below.

oss_crs/src/crs_compose.py
bmendonca3: @azchin addressed this in the latest push.

I moved the interpolation helper into `env_policy.py`, kept required unresolved envs as preflight failures, and added a warning for optional `additional_env` placeholders that reference unset host env vars but are not listed in `required_envs`.

Local verification passed with `uv run verify`.

azchin:
Tested with `crs-bug-finding-claude-code` and `compose-oauth.yaml`. LGTM

PR #239, opened 2026-05-28
fix: use user-scoped snapshot locks

### LINKED ISSUE #132: Snapshot lock file causes Permission denied when run by different user
## Bug

`oss-crs build-target` fails with `Permission denied` on the snapshot lock file when run by a different user than the one that created it.

```
Error: Create snapshot (address)
[Errno 13] Permission denied: '/tmp/oss-crs-snapshot-locks/snapshot-c37fcd4539fdb920.lock'
```

## Root Cause

`file_lock()` in `oss_crs/src/target.py` creates lock files under `/tmp/oss-crs-snapshot-locks/` with the default umask of the running user. The lock file is never deleted after use — it persists across runs.

If a previous run created the lock file as a different user (e.g., root via Docker/sudo), subsequent runs by a normal user get `EACCES` when trying to `open(lock_path, "w")`.

The `fcntl.flock` advisory lock itself is properly released in the `finally` block, so the file isn't "held" — it's just owned by the wrong user.

## Reproduction

1. Run `oss-crs build-target` as root (or via sudo)
2. Run `oss-crs build-target` again as a normal user for the same target
3. Second run fails with `Permission denied` on the lock file

## Workaround

```bash
sudo rm /tmp/oss-crs-snapshot-locks/snapshot-*.lock
```

## Possible Fixes

1. **Delete lock file after release** — add `lock_path.unlink(missing_ok=True)` in the `finally` block. Simple but introduces a small TOCTOU race between parallel runs.
2. **Create with permissive mode** — use `os.open(lock_path, os.O_CREAT | os.O_WRONLY, 0o666)` so any user can re-acquire the lock.
3. **Use user-scoped lock dir** — e.g., `$XDG_RUNTIME_DIR` instead of `/tmp`, so each user gets their own lock files.

bmendonca3:
## Summary
- move incremental snapshot lock files from the shared `/tmp/oss-crs-snapshot-locks` path to a per-user runtime/cache directory
- prefer `XDG_RUNTIME_DIR`, fall back to `XDG_CACHE_HOME`, then `~/.cache/oss-crs/snapshot-locks`
- add tests covering lock directory selection and `_snapshot_one` lock path usage

Fixes #132.

## Testing
- `uv run verify`

### DIFF
--- a/oss_crs/src/crs_compose.py
+++ b/oss_crs/src/crs_compose.py
@@ -24,6 +24,7 @@ from .utils import (
     log_success,
     log_warning,
     log_dim,
+    user_temporary_dir,
 )
 from .workdir import WorkDir
 from .cgroup import (
@@ -35,19 +36,6 @@ from .cgroup import (
 import docker
 import docker.errors
 import requests.exceptions
-
-
-def _snapshot_lock_dir() -> Path:
-    """Return a per-user lock directory for incremental snapshot creation."""
-    runtime_dir = os.environ.get("XDG_RUNTIME_DIR")
-    if runtime_dir:
-        return Path(runtime_dir) / "oss-crs" / "snapshot-locks"
-
-    cache_dir = os.environ.get("XDG_CACHE_HOME")
-    if cache_dir:
-        return Path(cache_dir) / "oss-crs" / "snapshot-locks"
-
-    return Path.home() / ".cache" / "oss-crs" / "snapshot-locks"
 
 
 class CRSCompose:
@@ -215,7 +203,7 @@ class CRSCompose:
 
         # File lock keyed on the content hash — serializes concurrent snapshot
         # creation for the same content across processes.
-        lock_dir = _snapshot_lock_dir()
+        lock_dir = user_temporary_dir() / "snapshot-locks"
         lock_path = lock_dir / f"snapshot-{content_key}.lock"
 
         with file_lock(lock_path):
--- a/oss_crs/src/crs_compose.py
+++ b/oss_crs/src/crs_compose.py
@@ -37,6 +37,19 @@ import docker.errors
 import requests.exceptions
 
 
+def _snapshot_lock_dir() -> Path:
+    """Return a per-user lock directory for incremental snapshot creation."""
+    runtime_dir = os.environ.get("XDG_RUNTIME_DIR")
+    if runtime_dir:
+        return Path(runtime_dir) / "oss-crs" / "snapshot-locks"
+
+    cache_dir = os.environ.get("XDG_CACHE_HOME")
+    if cache_dir:
+        return Path(cache_dir) / "oss-crs" / "snapshot-locks"
+
+    return Path.home() / ".cache" / "oss-crs" / "snapshot-locks"
+
+
 class CRSCompose:
     @classmethod
     def from_yaml_file(
@@ -202,7 +215,7 @@ class CRSCompose:
 
         # File lock keyed on the content hash — serializes concurrent snapshot
         # creation for the same content across processes.
-        lock_dir = Path("/tmp/oss-crs-snapshot-locks")
+        lock_dir = _snapshot_lock_dir()
         lock_path = lock_dir / f"snapshot-{content_key}.lock"
 
         with file_lock(lock_path):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,9 @@ stricter subset of Keep a Changelog).
   `_relative_repo_hint`).
 
 ### Fixed
+- Incremental snapshot locks now use a per-user runtime/cache directory instead
+  of a shared `/tmp/oss-crs-snapshot-locks` path, avoiding cross-user
+  permission failures from stale lock files.
 - The local run path now passes a `Path` compose-file object consistently into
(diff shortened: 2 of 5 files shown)


oss_crs/src/crs_compose.py
```diff
@@ -37,6 +37,19 @@
 import requests.exceptions
 
 
+def _snapshot_lock_dir() -> Path:
```
azchin: Can you change this to a more generic (and public) helper function that returns a per-user temp directory? I.e. remove the `snapshot-locks` suffix from each path.

We can reuse this helper later, move this function to `oss_crs/src/utils.py`

```suggestion
def user_temporary_dir() -> Path:
```

oss_crs/src/crs_compose.py
```diff
@@ -37,6 +37,19 @@
...
+    """Return a per-user lock directory for incremental snapshot creation."""
+    runtime_dir = os.environ.get("XDG_RUNTIME_DIR")
+    if runtime_dir:
+        return Path(runtime_dir) / "oss-crs" / "snapshot-locks"
```
azchin: For example,
```suggestion
        return Path(runtime_dir) / "oss-crs"
```

oss_crs/src/crs_compose.py
```diff
@@ -202,7 +215,7 @@ def _snapshot_one(
 
         # File lock keyed on the content hash — serializes concurrent snapshot
         # creation for the same content across processes.
-        lock_dir = Path("/tmp/oss-crs-snapshot-locks")
+        lock_dir = _snapshot_lock_dir()
```
azchin: And then append `"snapshot-locks"` to the per-user temp path here.
```suggestion
        lock_dir = user_temporary_dir() / "snapshot-locks"
```

oss_crs/src/crs_compose.py
```diff
@@ -37,6 +37,19 @@
...
 
+def _snapshot_lock_dir() -> Path:
+    """Return a per-user lock directory for incremental snapshot creation."""
+    runtime_dir = os.environ.get("XDG_RUNTIME_DIR")
+    if runtime_dir:
+        return Path(runtime_dir) / "oss-crs" / "snapshot-locks"
+
+    cache_dir = os.environ.get("XDG_CACHE_HOME")
+    if cache_dir:
+        return Path(cache_dir) / "oss-crs" / "snapshot-locks"
```
azchin:
```suggestion
        return Path(cache_dir) / "oss-crs"
```

oss_crs/src/crs_compose.py
```diff
@@ -37,6 +37,19 @@
...
+
+    return Path.home() / ".cache" / "oss-crs" / "snapshot-locks"
```
azchin:
```suggestion
    return Path.home() / ".cache" / "oss-crs" 
```

azchin (changes requested):
Hi Brian, thanks for your contribution and welcome to the OSS-CRS project! Please refer to my comments below. I will try to get to your other PR's by EOW.

oss_crs/src/crs_compose.py
```diff
@@ -37,6 +37,19 @@
 import requests.exceptions
 
 
+def _snapshot_lock_dir() -> Path:
+    """Return a per-user lock directory for incremental snapshot creation."""
+    runtime_dir = os.environ.get("XDG_RUNTIME_DIR")
```
azchin:
```suggestion
    user_directory = Path.home() / ".cache" # default to backup
    runtime_dir = os.environ.get("XDG_RUNTIME_DIR")
    cache_dir = os.environ.get("XDG_CACHE_HOME")
```

oss_crs/src/crs_compose.py
```diff
@@ -37,6 +37,19 @@
...
+    if runtime_dir:
+        return Path(runtime_dir) / "oss-crs" / "snapshot-locks"
+
+    cache_dir = os.environ.get("XDG_CACHE_HOME")
+    if cache_dir:
+        return Path(cache_dir) / "oss-crs" / "snapshot-locks"
+
+    return Path.home() / ".cache" / "oss-crs" / "snapshot-locks"
+
+
```
azchin:
```suggestion
    if runtime_dir and runtime_dir.is_dir():
        user_directory = runtime_dir
    elif cache_dir and cache_dir.is_dir():
        user_directory = cache_dir

    user_directory /= "oss-crs"
    return user_directory
```

azchin (changes requested):
~~Let's also make sure the directory exists in this helper.~~
EDIT: nevermind, this doesn't work for paths created with the helper by the caller. Let's just have the callers create the path.
EDIT 2: keeping the code suggestions to add checking for the directory's existence, and appending the magic string suffix cleaner

azchin:
Shoot, I realized this solution doesn't work well because the whole point of the lock is to prevent global races for Docker snapshotting (regardless of user).

Let's discuss this further on the issue #132 itself for better approaches.

Again thanks for your contribution and sorry to close it this far in.

bmendonca3:
Thanks Andrew, that makes sense. I was looking at the permission error and missed that the lock still needs to be shared across users for Docker snapshotting.

I’ll move this over to #132. Keeping the global lock path and making the lock file itself safe to reopen across users sounds like the better direction than using per-user lock dirs.

PR #240, opened 2026-05-28
fix: validate sidecar path identifiers

bmendonca3:
## Summary
- reject path-like `crs_name`, `harness_name`, and `rebuild_id` values at the sidecar API boundary
- return `400` before resolving artifact paths, checking harness files, launching subprocesses, or queueing build/test jobs
- add regression coverage for builder and runner sidecar traversal-style inputs

## Testing
- `uv run pytest oss_crs/tests/unit/test_runner_sidecar.py oss_crs/tests/unit/test_builder_sidecar.py`
- `uv run verify`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,8 @@ stricter subset of Keep a Changelog).
   `_relative_repo_hint`).
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


azchin (approved):
LGTM!

PR #241, opened 2026-05-28
fix(config): validate CRS entry names

bmendonca3:
## Summary
- validate CRS entry names at config load time with a conservative path-safe pattern
- reject traversal/path-like names, dots, spaces, uppercase letters, punctuation, and overlong names before they flow into paths, Docker tags, service aliases, and artifact directories
- add regression tests for accepted current-style names and rejected unsafe names

## Compatibility
- checked current `registry/*.yaml` names and `example/*/compose.yaml` CRS keys against the new pattern

## Testing
- `uv run pytest oss_crs/tests/unit/config/test_crs_compose.py`
- `uv run python - <<'PY' ...` compatibility check for registry/example CRS names
- `uv run verify`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,4 +85,5 @@ stricter subset of Keep a Changelog).
   the main local run path.
 
 ### Security
-- N/A
+- CRS entry names are now validated at config load time before being used in
+  paths, Docker tags, service aliases, and artifact directories.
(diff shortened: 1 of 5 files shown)


azchin (commented):
Thanks. I don't see the problem with CRS names with uppercase letters, could you expand the filter to allow them?

Also could you document this CRS name requirement in the CRS development guide?

bmendonca3:
Thanks, that makes sense. I updated the validator to allow uppercase letters while still rejecting path-like names, dots, spaces, punctuation outside hyphen/underscore, and overlong names. I also documented the CRS entry name format in the development guide and kept the config reference/tests in sync.

azchin (approved):
LGTM

PR #242, opened 2026-05-28
fix: make snapshot locks reusable across users

### LINKED ISSUE #132: Snapshot lock file causes Permission denied when run by different user

bmendonca3:
## Summary
- keep incremental snapshot locks in the shared `/tmp/oss-crs-snapshot-locks` namespace while creating the directory and lock files with shared permissions
- make shared permissions opt-in on `file_lock()` so repo checkout locks keep their existing behavior
- fall back to read-only `flock` for existing owner-write-only lock files created by older versions
- add regression tests for shared lock modes and snapshot lock usage

Fixes #132.

## Testing
- `uv run pytest oss_crs/tests/unit/test_file_lock.py oss_crs/tests/unit/test_incremental_snapshots.py`
- `uv run ruff check oss_crs/src/target.py oss_crs/src/crs_compose.py oss_crs/tests/unit/test_file_lock.py oss_crs/tests/unit/test_incremental_snapshots.py`
- `uv run ruff format --check oss_crs/src/target.py oss_crs/src/crs_compose.py oss_crs/tests/unit/test_file_lock.py oss_crs/tests/unit/test_incremental_snapshots.py`
- `uv run verify`

### DIFF
--- a/oss_crs/src/crs_compose.py
+++ b/oss_crs/src/crs_compose.py
@@ -205,7 +205,7 @@ class CRSCompose:
         lock_dir = Path("/tmp/oss-crs-snapshot-locks")
         lock_path = lock_dir / f"snapshot-{content_key}.lock"
 
-        with file_lock(lock_path):
+        with file_lock(lock_path, shared_permissions=True):
             # Re-check after acquiring lock — another process may have built it.
             try:
                 img = client.images.get(content_tag)
--- a/oss_crs/src/target.py
+++ b/oss_crs/src/target.py
@@ -2,6 +2,7 @@
 from pathlib import Path
 from typing import Optional
 import hashlib
+import os
 import posixpath
 import re
 import subprocess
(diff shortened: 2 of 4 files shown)


azchin (approved):
Looks good, I see the `/tmp/oss-crs-snapshot-locks` lock file with perm 666 after running `oss-crs build-target --incremental`

PR #319, opened 2026-07-14
Add support for building dependencies offline using nix

br32710:
## Summary

Describe what changed and why.

libCRS is now built during the prepare step using a Nix derivation and is stored as a docker image. During the build-target/run steps, the Nix store is copied into /nix/store which is then usable by the install script/run time images. Only the libCRS CLI tool is supported, not the python library.

## User Impact

- [ ] User-facing change
- [x] Internal-only change

If user-facing, describe CLI/API/config/docs impact and migration steps.

## Release Note / Changelog

- [ ] I updated `CHANGELOG.md` (`[Unreleased]`) for user-facing changes
- [X] No changelog entry needed (internal-only refactor/test/chore)

If this PR includes deprecation or breaking behavior, include:
- replacement path for users
- planned removal version/release window

## Validation

List tests/checks run and their results.

Successfully passes all unit tests.

## Checklist

- [X] I updated docs for behavior/config/CLI changes
- [X] I added/updated tests for behavior changes
- [X] I considered backward compatibility and migration impact

### DIFF
--- a/libCRS/flake.nix
+++ b/libCRS/flake.nix
@@ -1,0 +1,37 @@
+{
+  description = "libCRS — Nix-built Docker image providing libCRS + rsync";
+  inputs.nixpkgs.url =
+    "github:NixOS/nixpkgs/nixos-unstable";
+  outputs = { self, nixpkgs }:
+    let
+      forAll = f: nixpkgs.lib.genAttrs [ "x86_64-linux" "aarch64-linux" ]
+        (system: f nixpkgs.legacyPackages.${system});
+    in {
+      packages = forAll (pkgs: rec {
+        libCRS-pkg = pkgs.python3Packages.buildPythonPackage {
+          pname = "libcrs";
+          version = "0.1.0";
+          src = ./.;
+          pyproject = true;
+          build-system = [ pkgs.python3Packages.setuptools ];
+          dependencies = with pkgs.python3Packages; [ watchdog requests ];
+          doCheck = false;
+        };
+
+        # A single Nix closure containing Python (with libCRS installed) and
+        # rsync.  All three are reachable via <closure>/bin/{libCRS,rsync,python3}.
+        libcrs-runtime = pkgs.buildEnv {
+          name = "libcrs-runtime";
+          paths = [
+            (pkgs.python3.withPackages (_: [ libCRS-pkg ]))
+            pkgs.rsync
+          ];
+        };
+
+        # The runtime closure is consumed by libCRS/deps.Dockerfile, which
+        # runs `nix build .#libcrs-runtime` inside a nixos/nix container and
+        # copies the closure into the "oss-crs-deps" image via `docker build`.
+        default = libcrs-runtime;
+      });
+    };
+}
--- a/oss_crs/src/constants.py
+++ b/oss_crs/src/constants.py
@@ -30,6 +30,18 @@ BASE_RUNNER_IMAGE = "gcr.io/oss-fuzz-base/base-runner"
 LEGACY_BASE_OS_VERSION = "legacy"
 DEFAULT_BASE_RUNNER_TAG = "latest"
 
+# Pinned Nix builder image used to build the oss-crs-deps Docker image during
+# prepare. Passed as the NIX_BUILDER_IMAGE build-arg to libCRS/deps.Dockerfile,
+# where Nix runs inside an ephemeral Docker container; the host's own /nix is
+# never touched or required.
+NIX_BUILDER_IMAGE = "nixos/nix@sha256:e623d73af9cac82d1b50784c83e0cf2a4b83bfd2cfe8d5b67809a2fc94e043ac"  # v2.28.3
+
+# The constant Docker image name for the Nix-built libCRS+rsync dependencies
+# image. CRS builder Dockerfiles reference this via COPY --from=oss-crs-deps.
+# Built during prepare via `docker build` (see oss_crs/src/libcrs_nix.py and
+# libCRS/deps.Dockerfile) and tagged oss-crs-deps:latest.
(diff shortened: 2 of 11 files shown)


libCRS/flake.nix
```diff
@@ -0,0 +1,41 @@
+{
+  description = "libCRS — Nix-built Docker image providing libCRS + rsync";
+  # Pin nixpkgs to a full commit SHA rather than a branch name (e.g.
+  # nixos-unstable).  Resolving a branch requires a GitHub commits-API
+  # call on every build, which is rate-limited (HTTP 403) and defeats
+  # reproducibility.  A full revision is fetched as a tarball directly.
+  inputs.nixpkgs.url =
+    "github:NixOS/nixpkgs/e7a3ca8092b61ff85b6a45bf863ea2b2d6a661b3";
```
azchin: The more idiomatic thing is to set this to nixos-unstable, and then the `flake.lock` file pins the hash. Commit the `flake.lock`.

oss_crs/src/constants.py
```diff
@@ -30,6 +30,18 @@
 LEGACY_BASE_OS_VERSION = "legacy"
 DEFAULT_BASE_RUNNER_TAG = "latest"
 
+# Pinned Nix builder image used to build the oss-crs-deps Docker image during
+# prepare. Passed as the NIX_BUILDER_IMAGE build-arg to libCRS/deps.Dockerfile,
+# where Nix runs inside an ephemeral Docker container; the host's own /nix is
+# never touched or required.
+NIX_BUILDER_IMAGE = "nixos/nix:2.28.3"
```
azchin: Use checksum for image here, and add in-line comment for human-readable version. See https://github.com/ossf/oss-crs/blob/abdc6450ab50c858967a7d28450c9ac5ecf6bc2f/oss_crs/src/constants.py#L3

oss_crs/src/crs_compose.py
```diff
@@ -583,7 +583,24 @@ def _prepare_build_fetch_dir(
...
+        """Build the oss-crs-deps Docker image (libCRS + rsync via Nix).
+
+        This is a framework-level prepare step that runs once during prepare.
+        The image is built with ``docker build`` (Nix runs inside the build
+        stage) and tagged ``oss-crs-deps:latest``. CRS builder Dockerfiles can
+        then use ``COPY --from=oss-crs-deps`` to get libCRS and rsync without
+        any network access at build time.
+        """
+        from .libcrs_nix import build_deps_image
+        from .templates import renderer
```
azchin: Please make these global imports

oss_crs/src/libcrs_nix.py
```diff
@@ -0,0 +1,106 @@
...
+``docker build``.  Its build stage runs Nix inside a pinned ``nixos/nix``
+container to produce the ``libcrs-runtime`` closure, then assembles a
+minimal ``FROM scratch`` image tagged ``oss-crs-deps:latest``.
+
+The image contains the full ``libcrs-runtime`` Nix closure (so that
+``/nix/store`` is populated), together with two baked symlinks::
+
+    /usr/local/bin/libCRS  ->  /nix/store/<hash>-libcrs-runtime/bin/libCRS
+    /usr/local/bin/rsync   ->  /nix/store/<hash>-libcrs-runtime/bin/rsync
+
```
azchin, line 13: What about the libCRS Python API?

azchin (commented):
Sorry for the delay, I did not get the notification for this PR. Here are some quick comments, I will try running your branch using a real CRS later. And you can ignore the Kusari suggestion, since you're using the image as purely a file store.

azchin:
Please run `ruff format` and re-push. You can run the CI static verification locally with `uv run verify`

oss_crs/src/libcrs_nix.py
br32710: I was looking into the minimal set of changes we needed to get it to initially work: we could add support for the libCRS Python API but it requires a lot more changes for the CRS developers. Since the libCRS/rsync dependencies are only needed for Dockerfiles which are based on the target project image, and it seems like libCRS is only needed as a python library for runtime images, the only real problem occurs for run time images which are based on the target project. The only dockerfile for which this is a problem is lsp.Dockerfile in atlantis-multilang-wo-concolic, but it doesn't seem to use libCRS as a python library. If this is true, we can avoid immediately adding support for the libCRS python API.

br32710:
Thanks for the review, I updated the pull request based on the comments

oss_crs/src/libcrs_nix.py
br32710: It looks like libCRS is used as a python library in lsp.Dockerfile, but that would probably be resolved in a different pull request upstream in atlantis-multilang-wo-concolic by adding a dockerfile to build all the dependencies using nix since the libCRS library used is different from the one here. This image is meant for the oss-crs dependencies across multiple CRS's

azchin: That's a misnomer, atlantis-multilang (and other Atlantis CRSs) have an old "libCRS" library that was used for the competition, which is different from the OSS-CRS libCRS. `lsp.Dockerfile` does indeed install OSS-CRS's libCRS as CLI. I think what you're saying is fair, let's just accept that libCRS for now only works as a CLI in offline mode.

br32710: Great, I will update the docs to try to make it more clear that it only supports CLI

azchin: Ah, the agentic harness CRSs use the Python API (e.g. `crs-bug-finding-claude-code`)... I'll do some tinkering

azchin: I documented some ideas in #322, but probably the easiest option is to change the CRS to using `libCRS` CLI if it needs to be used offline.

PR #327, opened 2026-07-21
Add export and import subcommands

br32710:
## Summary

Describe what changed and why.

Added a export subcommand:
- Exports docker images pulled/built in the prepare phase and the CRS source code as a tar file, where the docker images are compressed using the zstd algorithm. Streaming is used to reduce peak file size.

Added an import subcommand:
- Restores OSS-CRS's state from a tar file generated using the export subcommand. Streaming is used to load the docker images.

## User Impact

- [X] User-facing change
- [ ] Internal-only change

If user-facing, describe CLI/API/config/docs impact and migration steps.

## Release Note / Changelog

- [x] I updated `CHANGELOG.md` (`[Unreleased]`) for user-facing changes
- [ ] No changelog entry needed (internal-only refactor/test/chore)

If this PR includes deprecation or breaking behavior, include:
- replacement path for users
- planned removal version/release window

## Validation

List tests/checks run and their results.

## Checklist

- [x] I updated docs for behavior/config/CLI changes
- [x] I added/updated tests for behavior changes
- [x] I considered backward compatibility and migration impact

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ stricter subset of Keep a Changelog).
 
 ### Changed
 
+- `oss-crs export` and `oss-crs import` commands — imports/exports docker images/CRS source code to transfer to another host.
 - Run-phase modules now default to `target_dependent: true`, so their images are built once per target during `build-target`. Set `target_dependent: false` for modules that can be built once during `prepare`.
 - `--offline` flag for all subcommands: disables git fetch
 - `oss-crs build-target` now validates `--bug-candidate`/`--bug-candidate-dir` the same way `oss-crs run` does: a missing path, or a directory passed to `--bug-candidate` (or a file passed to `--bug-candidate-dir`), fails before any container starts instead of being silently ignored.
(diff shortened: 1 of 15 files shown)


br32710:
Tested export/import on atlantis-multilang-wo-concolic.

Notes: --compose-file may not be necessary for the import command, could switch to gzip instead of zstd since gzip is more common, had problems with busybox tar since it doesn't seem to support the pax format output by python.

azchin:
`handle_import` stages into `<work-dir>/crs_compose/.import-XXXX/` and cleans up with rm_with_docker (shells out to `docker run --rm ... oss-crs-alpine:latest rm -rf`)

On a host that has never run prepare, that tag doesn't exist. Docker attempts a registry pull, gets pull access denied, exits 125, and rm_with_docker raises RuntimeError from the finally, replacing the pending return False.

The success path is saved only by ordering: _load_images installs oss-crs-alpine from the bundle before cleanup runs. Every early-exit path returns first — missing manifest (:811), platform mismatch (:821), _load_images failure (:830), extractall raising (:806).

Fix: use `shutil.rmtree(staging, ignore_errors=True)`. The staging dir is created by tempfile.mkdtemp in this process and holds only files this process extracted — all user-owned. rm_with_docker exists for root-owned container-written files, which don't occur here. (_restore_dir at :83 is different — it replaces a CRS source tree that may contain root-owned files, so it keeps a legitimate claim on the helper.)

Repro: `docker rmi oss-crs-alpine:latest`, then `oss-crs import --in <any-malformed>.tar --work-dir /tmp/wd`.

br32710:
Fair, I'll update that.

PR #361, opened 2026-08-18
feat(cli): add list-harnesses command

### LINKED ISSUE #70: Add CLI command to list runnable harnesses for a target project
## Motivation

When working with an OSS-Fuzz target project, there's no easy way to find out which harness binaries are available to run. The harnesses are only discoverable after building the target image (they live in `/out/` inside the docker image), and right now you have to manually inspect the container to find them.

A quick `crs-compose list-harnesses` command would save time, especially when exploring unfamiliar projects or when a project produces many harness binaries.

## Proposed behavior

```
crs-compose list-harnesses --target-proj-path <path> [--work-dir <dir>]
```

- If the target docker image is already built, list the harness binaries from `/out/` in the image.
- If the image hasn't been built yet, build it first (reuse existing build logic), then list.
- Output should be a simple list of harness binary names, one per line.

Something like:

```
$ crs-compose list-harnesses --target-proj-path ./projects/curl
curl_fuzzer_http
curl_fuzzer_ftp
curl_fuzzer_smtp
...
```

## Implementation notes

- The target image tag can be derived using `Target.get_docker_image_name()`.
- Listing `/out/` can be done with a quick `docker run --rm <image> ls /out/` or similar.
- Might want to filter out non-executable files or known non-harness files (e.g. `.options`, `.dict`).

HaikalE:
Closes #70.

## Summary

Adds a standalone `oss-crs list-harnesses --fuzz-proj-path <PATH>` command for building an OSS-Fuzz project and listing the runnable fuzz harnesses produced in `/out`.

The implementation follows the direction discussed in #70:

- builds through the project's OSS-Fuzz Docker/`compile` path
- caches `/out` and `/work` under `.oss-crs-workdir`
- invalidates the cache when project build inputs or repository HEAD changes
- filters support files and identifies runnable harnesses using language-specific heuristics
- keeps harness names on stdout while build/progress output goes to stderr

## Validation

Unit / static validation:

- 43 relevant tests passed
- Ruff check passed
- Ruff format check passed
- Pyright: 0 errors, 0 warnings
- `git diff --check` passed

Real OSS-Fuzz validation was performed against:

- `json-c` (C++)
- `libpng` (C; project without a local `build.sh`)
- `jackson-core` (JVM)
- `pyyaml` (Python)

Across these projects, 16 detected harnesses were verified runnable with OSS-Fuzz `base-runner test_one.py`.

I also verified:

- cold builds use the real OSS-Fuzz compile path
- warm invocations reuse cached `/out` artifacts without recompiling
- project-input changes invalidate the cache and rebuild
- cache access is serialized with a file lock
- stdout contains only harness names
- experimental sanitizer metadata in `project.yaml` is accepted

Harness detection remains heuristic. Projects using `FUZZING_ENGINE=none` may produce runnable executables without `LLVMFuzzerTestOneInput`; this limitation is documented in the README.


oss_crs/src/cli/list_harnesses.py
```diff
@@ -0,0 +1,391 @@
...
+        if not path.exists():
+            continue
+        try:
+            shutil.rmtree(path)
+        except PermissionError:
+            # OSS-Fuzz builds run as root and may leave nested directories that
+            # the host user cannot traverse or remove directly. Use the target
+            # image that was just built so this standalone command does not
+            # depend on OSS-CRS's internal Alpine helper image being present.
+            subprocess.run(
```
azchin: Can you replace this with `utils.rm_with_docker`? You can add a new parameter to that function if you want to specify a different image tag (i.e. OSS-Fuzz instead of Alpine)

oss_crs/src/cli/list_harnesses.py
```diff
@@ -0,0 +1,391 @@
...
+                    image_tag,
+                    "-rf",
+                    f"/data/{path.name}",
+                ],
+                check=True,
+                stdout=subprocess.DEVNULL,
+            )
+
+
+def handle_list_harnesses(args) -> bool:
```
azchin: Add a file lock when accessing the cache directory

oss_crs/src/cli/list_harnesses.py
```diff
@@ -0,0 +1,391 @@
...
+    return hashlib.sha256(encoded).hexdigest()
+
+
+def _build_sanitizer(target: Target) -> str:
+    # OSS-Fuzz's helper defaults JavaScript projects to an uninstrumented build.
+    return "none" if target.language == "javascript" else target.sanitizer
+
+
+def _compile_project(target: Target, image_tag: str, out_dir: Path) -> bool:
+    environment = {
```
azchin, line 120: Can we reuse `Target.get_target_env` and `OSS_FUZZ_TARGET_ENV` instead of redefining the dictionary here?

oss_crs/src/cli/list_harnesses.py
```diff
@@ -0,0 +1,391 @@
...
+        hasher.update(f"file:{relative.as_posix()}\0".encode())
+        hasher.update(str(stat.S_IMODE(path.stat().st_mode)).encode())
+        hasher.update(b"\0")
+        with path.open("rb") as stream:
+            for chunk in iter(lambda: stream.read(1024 * 1024), b""):
+                hasher.update(chunk)
+    return hasher.hexdigest()
+
+
+def _containing_repository_head(project_dir: Path) -> str | None:
```
azchin: Can we reuse `Target.get_repo_hash` instead of implementing a new function? Might have to extract it outside of the `Target` class as a library function instead of a class method.

oss_crs/src/cli/list_harnesses.py
```diff
@@ -0,0 +1,391 @@
...
+    """Return the HEAD of the checkout containing the OSS-Fuzz project."""
+    try:
+        repository = git.Repo(project_dir, search_parent_directories=True)
+        return repository.head.commit.hexsha
+    except (git.GitError, ValueError):
+        return None
+
+
+def _remote_repository_head(url: str | None) -> str | None:
```
azchin: Can you put this into `Target` as a method? Or similar to the `Target.get_repo_hash`, move it so it's more visible as a library API.

oss_crs/src/cli/list_harnesses.py
```diff
@@ -0,0 +1,391 @@
...
+    if not project_dir.is_dir():
+        _error(f"OSS-Fuzz project directory does not exist: {project_dir}")
+        return False
+    for required in ("Dockerfile", "build.sh"):
+        if not (project_dir / required).is_file():
+            _error(f"OSS-Fuzz project is missing {required}: {project_dir}")
+            return False
+
+    target = Target(args.work_dir, project_dir, None)
+    cache_dir = _cache_dir(args.work_dir, project_dir)
```
azchin, line 260: Add file lock when accessing the cache dir

oss_crs/src/cli/list_harnesses.py
```diff
@@ -0,0 +1,391 @@
...
+    )
+    parser.add_argument(
+        "--work-dir",
+        type=Path,
+        default=default_work_dir,
+        help="Working directory used to cache compiled harnesses",
+    )
+
+
+def _project_content_hash(project_dir: Path) -> str:
```
azchin: This is pretty similar to `Target.__get_proj_hash`, I would prefer to have the latter reused if possible. However, I prefer your current implementation since it's more robust to any changes in the project directory. Please move your implementation to `Target.__get_proj_hash`.

oss_crs/src/cli/list_harnesses.py
```diff
@@ -0,0 +1,391 @@
...
+    """Hash all project inputs that can affect an OSS-Fuzz build."""
+    hasher = hashlib.sha256()
+    for path in sorted(project_dir.rglob("*"), key=lambda item: item.as_posix()):
+        relative = path.relative_to(project_dir)
+        if ".git" in relative.parts or "__pycache__" in relative.parts:
```
azchin: Why is `__pycache__` here? Curious about what cases generate that in the project directory.

azchin (commented):
Thank you for the PR! I have comments primarily regarding reusing existing code. A lot of similar functionality exists in class methods (e.g. `Target`), so I'll leave it to your discretion whether some methods need to be moved out of the class to make it more flexible to reuse.

There are a few edge cases that fail the harness detection:
- `build.sh` being tracked in the target's repository and missing in the project directory
- `FUZZING_ENGINE` set to none, and the executable harness is missing the LLVMFuzzerTestOneInput marker (i.e. v8, jsc, spidermonkey, deno). This one is hard to handle systematically, just document it
- `project.yaml` fails pydantic validation because the sanitizer list contains KVPs (see below)
```yaml
sanitizers:
  - address
  - memory:
      experimental: true
```

HaikalE:
Addressed the review feedback in `a553cfa`.

The update reuses the existing `Target`/environment/hash and Docker cleanup APIs, adds cache locking, handles projects without a local `build.sh` and experimental sanitizer metadata, and documents the `FUZZING_ENGINE=none` limitation.

I also reran the relevant validation: 43 tests passed, Ruff/Pyright/diff checks are clean, and 16 harnesses across `json-c`, `libpng`, `jackson-core`, and `pyyaml` were verified with OSS-Fuzz `test_one.py`.

Thanks for the detailed review.

oss_crs/src/cli/list_harnesses.py
HaikalE: I couldn't find a concrete OSS-CRS/OSS-Fuzz case that requires excluding `__pycache__` here. The build runs in the container and shouldn't generate it in the host project input directory, so I removed the special-case exclusion.

If `__pycache__` is present in the project directory, it now contributes to the project hash like any other input.

azchin:
Looks good to me. The failing unit test is because `crs.py` imports enums from `TargetSanitizer` etc, which was changed in this PR.

Could you modify `oss_crs/src/config/crs.py` to explicitly define the default set of target configs? This isn't the cleanest way to handle, but fixing it is more of a design problem so I'll create a new issue. Add a TODO comment to mark.

In the meanwhile, this should fix the complaining unit tests.

```python
DEFAULT_SUPPORTED_SANITIZERS = frozenset({TargetSanitizer.ASAN, TargetSanitizer.MSAN, TargetSanitizer.UBSAN})
DEFAULT_SUPPORTED_ARCHITECTURES = frozenset({TargetArch.X86_64, TargetArch.I386})
DEFAULT_SUPPORTED_ENGINES = frozenset({
    FuzzingEngine.LIBFUZZER, FuzzingEngine.AFL, FuzzingEngine.HONGGFUZZ, FuzzingEngine.CENTIPEDE,
})


class SupportedTarget(BaseModel):
    mode: Set[TargetMode]
    language: Set[TargetLanguage]
    sanitizer: Set[TargetSanitizer] = Field(default_factory=lambda: set(DEFAULT_SUPPORTED_SANITIZERS))
    architecture: Set[TargetArch] = Field(default_factory=lambda: set(DEFAULT_SUPPORTED_ARCHITECTURES))
    fuzzing_engine: Set[FuzzingEngine] = Field(default_factory=lambda: set(DEFAULT_SUPPORTED_ENGINES))
```

HaikalE:
Addressed this in the latest commit. SupportedTarget now uses explicit default sanitizer, architecture, and fuzzing-engine sets, with a TODO for the broader design cleanup. The unit tests and static checks are clean.

